PR #11, opened 2015-09-18
Expose headers

azalan:
Expose the headers to enable framework users to use the classes in this framework. This PR only exposes those headers whose name starts with FB.
Also it would be nice to have an umbrella header but the FBSimulatorControl header is already taken by a class.


lawrencelomax:
I'm confident that the build will be fine, it just needs rebasing on the travis config I've just added.

PR #14, opened 2015-09-21
Guard against races in Simulator deletion

lawrencelomax:
When running allocation/deletion repeatedly, it's possible to expose what looks to be [`-[SimDeviceSet deleteDevice:error:]`](https://github.com/facebook/FBSimulatorControl/blob/master/PrivateHeaders/CoreSimulator/SimDeviceSet.h#L37) returning too early. 

I've managed to reproduce frequent failure of the new `-[FBSimulatorPoolAllocationTests testRemovesDeletedDeviceFromSet]` test, which can also be exposed in the startup path of `FBSimulatorControl`.

Instead of relying on the `SimDeviceSet` API doing the right thing, we can block on the existence of the newly-deleted Simulator in the DeviceSet and fail if this times out.


mmmulani:
how much does this delay deleting, on average? (like are we talking 0.5s or 5.0s)

lawrencelomax:
I haven't seen it take any longer than a few seconds on my MBP. Do you think the timeout should be more aggressive than 30s?

mmmulani:
no I think 30s is okay but maybe we should make `spinRunLoopWithTimeout` poll a bit more often? anyways, 👍🏾

lawrencelomax:
Oh yeah, polling every 1s is probably making a lot of waits take longer than they to, so I'll follow up with a change to that as well.

PR #15, opened 2015-09-21
Configurable prefix for Pool-Managed Simulators

lawrencelomax:
Adds an additional property to `FBSimulatorControlConfiguration` to allow for prefixes other than `E2E`. A prefix *must* be non-empty for the meantime, until it is possible to work this out with a separately managed `SimDeviceSet`.


mmmulani:
lgtm

PR #16, opened 2015-09-22
Fix batched-deletion SimDevices sticking around

lawrencelomax:
#14 waited on Simulators to actually be removed from the `SimDeviceSet` before returning from a deletion method, however this wasn't used in the batch deletion method `-[FBSimulatorPool deleteSimulators:withError:]. This PR fixes that.


mmmulani:
lgtm

PR #17, opened 2015-09-22
Increase runloop spin frequency

lawrencelomax:
We probably waste a lot of time waiting on conditions that resolve very quickly. Instead of waiting for this every second we can still be prudent in terms of system resources (by not checking the condition continuously) but also resolve conditions faster by setting this to 100ms instead


mmmulani:
lgtm

Issue #10, opened 2015-09-17
Question: getting notified when a test run finished


arpadzalan:
I was able to run an `xctest` using the FBSimulatorControl library by passing in the same environment and command line arguments as XCode does when it starts the tests.
I quickly realized that I have to build some kind of waiting mechanism before I terminate the simulator (which does the cleanup), but I haven't found anything on the API which could help me achieving this.

This is what I have tried:
- I tried to poll the `session.state.runningProcesses` to see whether the process is still there, but it seems that it is not removed from there.
- I realized that the app (with the injected tests) runs in a separate os process and I can get the process ID for it. With this information I can run `ps` to check whether the process is still alive (I could not get `repeatedlyRunCommand:withError:untilTrue:` working in `FBTaskExecutor` because of a validation timeout, but probably this direction could work).

I wonder if there is a better way to get notified on the end of the text execution.

lawrencelomax:
How do you launch the `xctest` process? I'm assuming you do this not by using `FBAgentLaunchConfiguration` but by using a task directly? There are notifications for processes that are launched by `FBSimulatorControlSessionInteraction`s such as `FBSimulatorSessionAgentProcessDidTerminateNotification`. All of the data that relates to simulator state is a product of interactions with the session, so any processes that aren't launched via `FBSimulatorControl` won't appear there.

I do have some changes that I need to upstream that will look at `launchd_sim` and the subprocesses of `launchd_sim` to determine if things like `backboardd` have crashed. Is the process that you care about a child of `launchd_sim`? If that's the case then you could potentially poll for those processes, or could provide a mechanism for adding a notification when a specific process that you care about, dies. 

Or, could you achieve this by using `FBDispatchSourceNotifier` on a specific PID?

azalan:
I have used `FBApplicationLaunchConfiguration` to launch the tests (by setting the environment variables and command line arguments just like XCode does). I tried to listen to `FBSimulatorSessionApplicationProcessDidTerminateNotification`, look into the runningApplications property, FBDispatchSourceNotifier and calling `ps` periodically. Only the last option worked.

I have created a project to test-drive this: https://github.com/azalan/SimCtrlCLI.

Could you give me a hint on what have I missed?

anshu10165:
@arpadzalan : I got the error below after running the same.
fatal error: 
      'FBSimulatorControl/FBDispatchSourceNotifier.h' file not found
[...]

PR #18, opened 2015-09-22
Add test for unexpected application termination notifications

lawrencelomax:
In #10 @arpadzalan was looking for a way to assert that some process dies in order to terminate the test. This adds a test for an unexpected Application Notification.


mmmulani:
lgtm

PR #21, opened 2015-09-23
Managed & Unmanaged FBSimulator

lawrencelomax:
Extracts out `FBSimulator` simulator behavior to allow all Simulators to be wrapped with the `FBSimulator` conveniences that are relevant, regardless of whether they are managed or unmanaged.

A Concreted subclass `FBManagedSimulator` deals with allocation/freeing of these Simulators. This improves a lot of the inflation of `SimDevice` to `FBSimulator` and means that `FBSimulatorPool` spends a lot less time dealing with heterogeneous objects.

This also means that expensive information about a simulator (for example the construction of processes that have been created through `launchd_sim`) can be done lazily and will apply to any simulator, not just 'managed' ones.

### DIFF
--- a/FBSimulatorControl/Management/FBSimulator+Private.h
+++ b/FBSimulatorControl/Management/FBSimulator+Private.h
@@ -9,12 +9,21 @@
 
 #import <FBSimulatorControl/FBSimulator.h>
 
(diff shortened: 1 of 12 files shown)


mmmulani:
lgtm

PR #22, opened 2015-09-23
Add Scale to Simulator Configuration

lawrencelomax:
When launching Simulators, it can be useful to set the scale to a more appropriate one depending on the desktop resolution of the machine launching the Simulator. 

This PR adds this into the configuration, with a default value and the necessary plumbing to fix this value via the Command Line to Simulator.app.


mmmulani:
lgtm

PR #23, opened 2015-09-23
Add Window Tiling

lawrencelomax:
Using the Mac OS X Accessibility APIs, it's possible to tile the windows so that they are in the foreground and do not overlap each other. This makes observing the windows in an automated fashion substantially easier.

Adds a test to this effect.

- [x] Figure out how to find the best non-occluded fit in a display
- [x] How to get this running on Travis-CI given that approval for accessibility is required.
- [x] Fix Build-Break
- [x] Tile in an `FBSimulatorSessionInteraction`
- [x] `CFTypeRef`s require releasing.


mmmulani:
lgtm

PR #25, opened 2015-09-23
Add inference of Simulator Process Identifiers

lawrencelomax:
The Process Identifier of a `FBSimulator` can be inferred from the arguments of the launched Simulator process. This allows `FBSimulatorControl` to use the PID to find the Windows of Simulator Applications that *haven't* been launched by the *current `FBSimulatorControl` process*.

Also added a test to confirm this.


mmmulani:
(Y)

PR #26, opened 2015-09-23
Composable NSPredicates for finding Simulators

lawrencelomax:
Extracts out the filtering of the `allSimulators` property of `FBSimulatorPool` into discrete `NSPredicate` objects in the `FBSimulatorPredicates` object. Thanks to `NSCompoundPredicate` it should become easier to define more complex filtering of Simulators.

By using #25 it should be possible to get a collection of Process Identifiers of *any* `FBSimulatorControl` launched Simulator, regardless of the process that it was launched from.


lawrencelomax:
Closing this since it's not possible to amend the target-branch of a Pull Request.

PR #30, opened 2015-09-24
Video Recording

lawrencelomax:
Builds on the work #29 and #23 to use `AVCaptureScreenInput` to record a video of multiple Simulators at once, to separate files.

- [x] Figure out why `Error Domain=AVFoundationErrorDomain Code=-11809 "Recording Stopped"` happens when the Simulator boots.


lawrencelomax:
The error code for `11809` is `AVErrorMediaChanged`, so the `Simulator.app` must be doing something that caused `AVFoundation` to suspend/cancel, recording. 

Recording is fine if it starts *after* the Simulator has booted to Springboard. In a single-sim environment this would be fine, but when booting two or more simulators at once, a booted Simulator will cancel recording for all Simulators at once. It should be possible to resume to the video, but this seems pretty terrible too as there will be gaps in the video.

lawrencelomax:
This looks to be a change in the Xcode 7 `Simulator.app`. Running an iOS 8 Simulator on Xcode 6.4 doesn't exhibit the `AVErrorMediaChanged` behavior. I guess this makes some sense, given that Simulator booting appears to be a bit different; there are progress bars before springboard is visible.

lawrencelomax:
The termination of the `AVCaptureMovieFileOutput` looks to be caused by a change in the `AVCaptureScreenInput`'s Pixel format. Before the Simulator has finished booting it is apparently '0000'. When the Simulator boots a `AVCaptureInputPortFormatDescriptionDidChangeNotification` will fire and the [Session re-starts](https://gist.github.com/lawrencelomax/feddd6faa45857ff7f2e). The re-starting session will cause the output to file to finish.

It would be really awesome if we could just use `AVCaptureMovieFileOutput` rather than having to resort to `AVCaptureVideoDataOutput`. I'm not entirely sure how this can be done, since it doesn't look like the pixel format can be changed for `AVCaptureMovieFileOutput`.

lawrencelomax:
There's a reasonable and non-terrible solution that involves deferring the start of the `AVCaptureMovieFileOutput` until after the `AVCaptureSessionDidStartRunningNotification` has started, which co-incides with the start of the Simulator. When a Capture Session has started, it will not be interrupted by other Simulators starting, so it will be fine in a multisim environment.

lawrencelomax:
Merging this now. Will look at a longer-term solution for the tests-on-travis later.

PR #33, opened 2015-09-30
Add Window Tiling Strategies

lawrencelomax:
There's more than one way of tiling windows within the Display bounds, so this should be pulled out into a strategy that defines the interface for interacting with other windows. 

For example, it may be possible to guarantee a maximum number of simulators, so offsetting in regions of a display is good enough.

Later, it may make sense to be able to compose these strategies. For example, there could be multiple tests in parallel with multiple simulators per-test. In this case the tests could be tiled horizontally and vertical tiling for the simulators within a test.


mmmulani:
lgtm

PR #34, opened 2015-09-30
Base XCTestCase

lawrencelomax:
Moves a lot of the duplicated code for creating `FBSimulatorControl` to a `XCTestCase` subclass. Drastically reduces the amount of boiler plate and provides a common way of cleaning up `FBSimulator` instances after they have been allocated by a Test Case.


mmmulani:
lgtm

PR #38, opened 2015-10-02
Remove Pods Dependencies & OCMock

lawrencelomax:
`OCMock` is only needed in a few tests, we can replace these with doubles and everything is a-ok. Also removes the need for a Podfile, so no additional steps are needed to run the tests.


mmmulani:
damn you @jspahrsummers!

lgtm though

PR #41, opened 2015-10-05
Support Custom SimDeviceSets

lawrencelomax:
It's possible to change `Simulator.app` to use a different device set by passing `-DeviceSetPath` as an arg to the launched `Simulator`. This will be vastly preferable to using Prefixes to isolate simulators between processes.

cc @mmmulani @ExtremeMan

### DIFF
--- a/FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.m
+++ b/FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.m
@@ -16,6 +16,7 @@ NSString *const FBSimulatorControlConfigurationDefaultNamePrefix = @"E2E";
 @interface FBSimulatorControlConfiguration ()
 
 @property (nonatomic, copy, readwrite) FBSimulatorApplication *simulatorApplication;
+@property (nonatomic, copy, readwrite) NSString *deviceSetPath;
 @property (nonatomic, copy, readwrite) NSString *namePrefix;
 @property (nonatomic, assign, readwrite) NSInteger bucketID;
 @property (nonatomic, assign, readwrite) FBSimulatorManagementOptions options;
@@ -24,13 +25,18 @@ NSString *const FBSimulatorControlConfigurationDefaultNamePrefix = @"E2E";
 
 @implementation FBSimulatorControlConfiguration
 
-+ (instancetype)configurationWithSimulatorApplication:(FBSimulatorApplication *)simulatorApplication namePrefix:(NSString *)namePrefix bucket:(NSInteger)bucketID options:(FBSimulatorManagementOptions)options
++ (instancetype)configurationWithSimulatorApplication:(FBSimulatorApplication *)simulatorApplication
+                                        deviceSetPath:(NSString *)deviceSetPath
+                                           namePrefix:(NSString *)namePrefix
+                                               bucket:(NSInteger)bucketID
+                                              options:(FBSimulatorManagementOptions)options
 {
   NSParameterAssert(simulatorApplication);
   NSParameterAssert(bucketID >= 0);
 
   FBSimulatorControlConfiguration *configuration = [self new];
   configuration.simulatorApplication = simulatorApplication;
+  configuration.deviceSetPath = deviceSetPath;
   configuration.namePrefix = namePrefix.length > 0 ? namePrefix : FBSimulatorControlConfigurationDefaultNamePrefix;
   configuration.bucketID = bucketID;
   configuration.options = options;
@@ -41,6 +47,7 @@ NSString *const FBSimulatorControlConfigurationDefaultNamePrefix = @"E2E";
 {
   return [self.class
     configurationWithSimulatorApplication:self.simulatorApplication
+    deviceSetPath:self.deviceSetPath
     namePrefix:self.namePrefix
     bucket:self.bucketID
     options:self.options];
@@ -48,7 +55,7 @@ NSString *const FBSimulatorControlConfigurationDefaultNamePrefix = @"E2E";
 
 - (NSUInteger)hash
 {
-  return self.simulatorApplication.hash | self.namePrefix.hash | self.bucketID | self.options;
+  return self.simulatorApplication.hash | self.deviceSetPath.hash | self.namePrefix.hash | self.bucketID | self.options;
 }
 
 - (BOOL)isEqual:(FBSimulatorControlConfiguration *)object
@@ -57,6 +64,7 @@ NSString *const FBSimulatorControlConfigurationDefaultNamePrefix = @"E2E";
     return NO;
   }
   return [self.simulatorApplication isEqual:object.simulatorApplication] &&
+         ((self.deviceSetPath == nil && object.deviceSetPath == nil) || [self.deviceSetPath isEqual:object.deviceSetPath]) &&
          [self.namePrefix isEqualToString:object.namePrefix] &&
          self.bucketID == object.bucketID &&
          self.options == object.options;
@@ -65,7 +73,8 @@ NSString *const FBSimulatorControlConfigurationDefaultNamePrefix = @"E2E";
 - (NSString *)description
 {
   return [NSString stringWithFormat:
-    @"Pool Config | Sim App %@ | Prefix %@ | Bucket Id %ld | Options %ld",
+    @"Pool Config | Set Path %@ | Sim App %@ | Prefix %@ | Bucket Id %ld | Options %ld",
+    self.deviceSetPath,
     self.simulatorApplication,
     self.namePrefix,
     self.bucketID,
--- a/FBSimulatorControl/Management/FBSimulatorControl.m
+++ b/FBSimulatorControl/Management/FBSimulatorControl.m
@@ -52,7 +52,7 @@
   }
 
   _configuration = configuration;
-  _simulatorPool = [FBSimulatorPool poolWithConfiguration:configuration deviceSet:SimDeviceSet.defaultSet];
+  _simulatorPool = [FBSimulatorControl poolForConfiguration:configuration];
   return self;
 }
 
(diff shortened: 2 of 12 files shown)


mmmulani:
tests red?

lawrencelomax:
The test is failing on pre-Xcode 7. I've updated the tests to take this into account and skip in the Custom Device Set case. 

We can get probably get this working for Xcode 6 by injecting and swizzling the Simulator app process (which we can't do for Xcode 7 because `amfid`). I am happy to do this later (if at all) for Xcode 6, if you agree?

mmmulani:
ahh, this lgtm then

I'm happy to pass on Xcode 6 support as long as we can get as far back as iOS 7 running under Xcode 7. (which should be possible if we copy the simulator SDKs over)

FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.m
```diff
@@ -48,7 +55,7 @@ - (instancetype)copyWithZone:(NSZone *)zone
 
 - (NSUInteger)hash
 {
-  return self.simulatorApplication.hash | self.namePrefix.hash | self.bucketID | self.options;
+  return self.simulatorApplication.hash | self.deviceSetPath.hash | self.namePrefix.hash | self.bucketID | self.options;
```
ExtremeMan: Interesting. I usually do `[_ns_object_1 hash] ^ [_ns_object_2 hash]`.

FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.m
```diff
@@ -57,6 +64,7 @@ - (BOOL)isEqual:(FBSimulatorControlConfiguration *)object
     return NO;
   }
   return [self.simulatorApplication isEqual:object.simulatorApplication] &&
+         ((self.deviceSetPath == nil && object.deviceSetPath == nil) || [self.deviceSetPath isEqual:object.deviceSetPath]) &&
```
ExtremeMan: Could be simplified to `(self.deviceSetPath == object.deviceSetPath) || [.. isEqual:]`

FBSimulatorControl/Management/FBSimulatorControl.m
```diff
@@ -115,4 +115,16 @@ - (BOOL)firstRunPreconditionsWithError:(NSError **)error
   return YES;
 }
 
++ (FBSimulatorPool *)poolForConfiguration:(FBSimulatorControlConfiguration *)configuration
+{
+  return [FBSimulatorPool poolWithConfiguration:configuration deviceSet:[self deviceSetForConfiguration:configuration]];
+}
+
++ (SimDeviceSet *)deviceSetForConfiguration:(FBSimulatorControlConfiguration *)configuration
```
ExtremeMan: Why don't put this into `FBSimulatorControlConfiguration` as a property?

FBSimulatorControl/Management/FBSimulatorControl.m
```diff
@@ -115,4 +115,16 @@ - (BOOL)firstRunPreconditionsWithError:(NSError **)error
   return YES;
 }
 
++ (FBSimulatorPool *)poolForConfiguration:(FBSimulatorControlConfiguration *)configuration
+{
+  return [FBSimulatorPool poolWithConfiguration:configuration deviceSet:[self deviceSetForConfiguration:configuration]];
```
ExtremeMan: Why do you need to pass device set which is taken from configuration which is already passed as an argument? Just read it from configuration in constructor?

FBSimulatorControl/Session/FBSimulatorSessionInteraction.m
```diff
@@ -49,15 +50,18 @@ - (instancetype)bootSimulator
...
+      simulator.configuration.lastScaleCommandLineSwitch, simulator.configuration.scaleString,
+    ]];
+    if (simulator.pool.configuration.deviceSetPath) {
+      [arguments addObjectsFromArray:@[@"-DeviceSetPath", simulator.pool.configuration.deviceSetPath]];
+    }
 
     id<FBTask> task = [FBTaskExecutor.sharedInstance
       taskWithLaunchPath:simulator.simulatorApplication.binary.path
-      arguments:arguments];
+      arguments:[arguments copy]];
```
ExtremeMan: `arguments` isn't used anywhere else - no need to copy.

FBSimulatorControl/Session/FBSimulatorSessionInteraction.m
```diff
@@ -49,15 +50,18 @@ - (instancetype)bootSimulator
...
 
   return [self interact:^ BOOL (NSError **error) {
-    NSArray *arguments = @[@"--args",
+    NSMutableArray *arguments = [NSMutableArray arrayWithArray:@[@"--args",
       @"-CurrentDeviceUDID", simulator.udid,
       @"-ConnectHardwareKeyboard", @"0",
-      simulator.configuration.lastScaleCommandLineSwitch, simulator.configuration.scaleString
-    ];
+      simulator.configuration.lastScaleCommandLineSwitch, simulator.configuration.scaleString,
+    ]];
```
ExtremeMan: I like to use `[@[] mutableCopy]` :P

FBSimulatorControlTests/Tests/FBSimulatorControlSimulatorLaunchTests.m
```diff
@@ -22,6 +22,7 @@
 #import <FBSimulatorControl/FBSimulatorSessionLifecycle.h>
 #import <FBSimulatorControl/FBSimulatorSessionState+Queries.h>
 #import <FBSimulatorControl/FBSimulatorSessionState.h>
+#import <FBSimulatorControl/FBSimulatorControlStaticConfiguration.h>
```
ExtremeMan: Headers aren't ordered =)

FBSimulatorControlTests/Tests/FBSimulatorControlSimulatorLaunchTests.m
```diff
@@ -137,3 +147,49 @@ - (void)testLaunchesMultipleSimulatorsConcurrently
...
+
+- (NSString *)deviceSetPath
+{
+  return [NSTemporaryDirectory() stringByAppendingPathComponent:@"FBSimulatorControlSimulatorLaunchTests_CustomSet"];
+}
+
+- (void)testLaunchesSingleSimulator
+{
+  if (!FBSimulatorControlStaticConfiguration.supportsCustomDeviceSets) {
+    NSLog(@"-[%@ %@] can't run as Custom Device Sets are not supported for this version of Xcode", NSStringFromClass(self.class), NSStringFromSelector(_cmd));
```
ExtremeMan: Shouldn't we log this when the method is actually attempted to be used on non-supported platform rather then in tests?

ExtremeMan:
Added some comments =) Feel free to ignore because they are mostly nits and I don't know what is preferred style here.

FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.m
lawrencelomax: `xor` makes a lot more sense I think. Will do this everywhere for all the hashes we have.

FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.m
lawrencelomax: Will follow this up for all value objects in another PR.

FBSimulatorControl/Session/FBSimulatorSessionInteraction.m
lawrencelomax: I don't have particularly strong feels about this but I usually go for immutability if performance isn't a concern.

FBSimulatorControl/Management/FBSimulatorControl.m
lawrencelomax: I want to change this in another PR that changes the startup path so that preconditions can be met ahead of time. `FBSimulatorControl` creates the `DeviceSet` since it is fail-able if for instance the directory isn't writable. A Pool can only be constructed if a `SimDeviceSet` can.

FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.m
michelf: Actually `[_ns_object_1 hash] ^ [_ns_object_2 hash]` will always be zero if the two objects are equal. I wouldn't recommend that in general because someone could easily exploit that to slow things down (probably not in this specific case though). But multiply the second hash by an arbitrary factor before XORing and it becomes pretty good.

Doing plain `[_ns_object_1 hash] | [_ns_object_2 hash]` will skew things toward the higher numbers (because bits can never return to zero). So hashes will tend to clash more than using XOR, especially if you are ORing a lot of them.

FBSimulatorControl/Management/FBSimulatorControl.m
lawrencelomax: It's called in the constructor and I don't see the need for lazy instantation of a property that will just be immediately instantiated. It also means that the when the initializer has returned control to the caller, it is effectively immutable.

FBSimulatorControlTests/Tests/FBSimulatorControlSimulatorLaunchTests.m
lawrencelomax: 👍

FBSimulatorControlTests/Tests/FBSimulatorControlSimulatorLaunchTests.m
lawrencelomax: Good point, I'll add an error condition to `bootSimulator` when a custom device set is attempted to be booted. It should be possible to manipulate a custom device set from Xcode 6, but booting a custom set isn't supported without a shim.

PR #42, opened 2015-10-05
Run Travis tests on Xcode 7

lawrencelomax:


lawrencelomax:
Looks like there could be some raceyness in the killing of all simulators in the pool. There's now a much better way of doing this now custom device sets are supported. It's also probably not essential for the pkill to succeed, if it fails the shutdown may be fine.

lawrencelomax:
I'm now leaning more to this being a permissions issue, if it's failing on the `kill` this consistently it's very unlikely to be a race, will have to do some more digging to find if something has changed in the travis environment.

lawrencelomax:
It's possible that this could be fixed by switching to using `kill` directly and avoiding using `pgrep`

lawrencelomax:
OMG

PR #43, opened 2015-10-06
Remove Managed/Unmanaged Simulator Distinction

lawrencelomax:
`SimDeviceSets` are a far superior way of ensuring that Simulators are segregated and put elsewhere. Removing the concept of a 'Managed' Simulator vastly simplifies the allocation process and means that the default device set doesn't get littered with Simulators. This will also allow other tools to use `FBSimulatorControl` to manipulate Simulators that were created externally from `FBSimulatorControl`


mmmulani:
lgtm

PR #53, opened 2015-10-22
Add Log Fetching

lawrencelomax:
`FBWritableLog` is a wrapper around any Log or piece of diagnostic data, be it a String, Binary Data or a File. It lazily transforms between these kinds of data and provides reasonable defaults for names and the ability to cache in-memory data to files so they can be read by other processes.

`FBSimulatorLogs` / `FBSimulatorSessionLogs` are classes that allow convenient access of logs. At the moment it just includes support for Crash Logs, but this can be extended in the future. The current method of fetching Crash Logs is to use substring matching on the parent process `launchd_sim` matching the expected `launchd_sim` process identifier.

`Shimulator` is a Sean Connery inspired shim for Simulator subprocesses that currently is used to force a crash in an arbitrary Simulator subprocess. At the moment this means sending an unrecognized selector to `NSFileManager`. A test validating that the crash log can be fetched as a result of a forced crash has also been added. `Shimulator` repurposed for other code-injection in the future.

A few other things:
- There's now a simple way of adding a dylib to load on process launches
- There's now an helper method for concurrent `filter`/`map` of arrays adapted from the `FBSimulatorApplication` implementation. Substring matching on a folder of crash logs is comparatively slow and this helps out.


mmmulani:
lgtm. in the future you might want to support being able to inject multiple dylibs but one is good for now

PR #56, opened 2015-10-23
Add Video Upload

hpique:
Adds support for adding videos to the Photos app.  

This is achieved by using the Shimulator to inject code into Photos app that adds the videos. The video paths are taken from the environment variable `SHIMULATOR_UPLOAD_VIDEO` that contains a colon (`:`) separated list of paths.

This is unfortunately not possible with CoreSimulator and very impractical by manipulating the file system directly.


lawrencelomax:
:tropical_drink:

PR #63, opened 2015-10-27
Improve robustness of killAll in DeviceSet

lawrencelomax:
Instead of building up a massive list of simulators to kill using `grep -e` based upon the contents of the `DeviceSet` it's easier to just kill all using the path to the `SimDeviceSet` in the params as this will always be passed. The same applies to Simulators launched without a DeviceSet, which just searches based upon the existence of `CurrentDeviceUDID`.

### DIFF
--- a/FBSimulatorControl/Management/FBSimulatorPool.m
+++ b/FBSimulatorControl/Management/FBSimulatorPool.m
@@ -88,6 +88,7 @@ static NSTimeInterval const FBSimulatorPoolDefaultWait = 30.0;
 {
   [self.allocatedUDIDs removeObject:simulator.udid];
 
+  // Killing is a pre-requesite for deleting/erasing
   NSError *innerError = nil;
   if (![self killSimulators:@[simulator] withError:&innerError]) {
     return [FBSimulatorError failBoolWithError:innerError description:@"Failed to Free Device in Killing Device" errorOut:error];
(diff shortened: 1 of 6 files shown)


FBSimulatorControl/Management/FBSimulatorPool.m
```diff
@@ -373,7 +382,7 @@ - (BOOL)blanketKillSimulatorsFromDifferentXcodeVersion:(NSError **)error
   // This command kills all Simulator.app binaries that *don't* match the current Simulator Binary Path.
   NSString *simulatorBinaryPath = self.configuration.simulatorApplication.binary.path;
   NSString *command = [NSString stringWithFormat:
-    @"pgrep -lf Simulator.app | grep -v %@ | awk '{print $1}' | xargs kill",
```
marekcirkos: Doesn't look right. Are you sure?

lawrencelomax: This string was prepended to [the string in `blanketKillSimulatorAppsPidProducingCommand:`](https://github.com/facebook/FBSimulatorControl/pull/63/files#diff-2a718c5690bba7ddc88dbb1d1c8dec3eR416) so you'd end up with `| xargs kill | xargs kill` otherwise. Verified this with the debugger.

PR #64, opened 2015-10-28
Figure out the Application Home Directory with launchctl

lawrencelomax:
Adds `pathToApplicationHome:` that calls `launchctl` on the Simulator in order to obtain the Home of the application from the `procinfo` of the Application. The Application Home can currently only be inferred if the Application is running, as `CoreSimulator` doesn't have this API


lawrencelomax:
Using `launchctl` to do this no longer makes much sense given that `FBProcessInfo` now has the environment of all processes launched by `launchd_sim`. It will be much easier to do it by filtering objects than some crazy shell command.

PR #67, opened 2015-11-19
Swift framework friendly

gblotter:
Gave the framework an Umbrella-Headers.h file, and created a module.modulemap file so that it can be exported as a Dynamic library and used in Swift applications.


lawrencelomax:
This is awesome! Hilariously I had something similar in the `fbsimctl` branch that I've now exported in #69. Could you rebase-on/merge #69 onto your branch, then I can merge your changes after?

gblotter:
@lawrencelomax So... I'm a bit confused.  Your code in #69 essentially does the same thing that my proposed changes would.  The main difference being that I tweaked the xcconfig settings to define it as a module (as specified in [Apple's Docs](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/BuildingCocoaApps/MixandMatch.html#//apple_ref/doc/uid/TP40014216-CH10-ID134) ).

So... what is it that you're looking for out of my code?  I'm willing to do whatever... just wanted to understand.

gblotter:
@lawrencelomax Took my best guess as to what you were wanting.  Let me know if you wanted something different.

lawrencelomax:
That's what I was looking for :+1:

PR #68, opened 2015-11-20
Add FBSimulatorTerminationStrategy

lawrencelomax:
Terminating Simulators is Hard™. This pulls out the nuts-and-bolts to a separate `FBSimulatorTerminationStrategy` class. The immediate effect is that `FBSimulatorPool` has a smaller footprint. Longer-term I'd like to look at using `sysctl` and `NSWorkspace` to terminate simulators in a more reasonable way.


mmmulani:
lgtm

PR #72, opened 2015-11-26
Add FBProcessQuery

lawrencelomax:
Adds `FBProcessQuery` which is a wrapper around the somewhat-complicated usage of `libproc` and `sysctl`. 

`libproc` is used as a simple interface to accessing basic information about running processes and the files that they access. `sysctl` is used for it's very handy `KERN_PROCARGS2` call, which allows the caller to access information about the launch path, environment and args of any process that the current user has permission to look at.

This class will enable `FBSimulatorControl` to:
- Kill off the janky `pgrep`ing that requires a subshell for simple tasks
- Infer the process identifier of arbitrarily launched simulators.
- Inspect the launch environment of all `launchd_sim` subprocesses.


mmmulani:
awesome! lgtm

marekcirkos:
lgtm

PR #73, opened 2015-11-26
Add option to enable CoreSimulator logging

mmmulani:
Added this as a C method because I couldn't find any good class that made sense for this.
Put it in the FBSimulatorPool header because that seems like the best place for it.


lawrencelomax:
Woo-yay

PR #75, opened 2015-11-27
Build fbsimctl on Travis

lawrencelomax:
Now that we have Xcode 7.1 magically working on Travis, we can build `fbsimctl` there to make sure that we're not breaking `fbsimctl` changes with Framework changes.


mmmulani:
lgtm

lawrencelomax:
Looks like it was that the `FBSimulatorControl.framework` dependency of `fbsimctl` needed to be 'Relative to Build Products'

PR #76, opened 2015-11-27
More performant search for launchd_sim

lawrencelomax:
Searching for `launchd_sim` by using paths is not performant and far too complicated than it needs to be. Finding the open processes for a file path is also orders of magnitude slower than getting simulator by launch path. It's also possible to optimise `processesWithLaunchPathSubstring:` later since `libproc` can provide all of them without the need for multiple syscalls.

Fortunately there are plenty of arguments and environment variables in the `launchd_sim` process that do contain the UDID of the Simulator, so we can use those instead.


mmmulani:
awesome

PR #81, opened 2015-12-01
Add FBSimulatorLaunchInfo

lawrencelomax:
Adds `FBSimulatorLaunchInfo` a class that has knowledge about launched `FBSimulator`s, regardless of how and when they were launched. This means that the process info for any `FBSimulator` can be obtained. Also adds self-updating of the `FBSimulator` instance to respond to application termination notifications. The absence of a `FBSimulatorLaunchInfo` object on a `FBSimulator` is also an indication that the Simulator has not been launched. This also paves the way for making `FBSimulatorSession` better.


mmmulani:
lgtm

PR #82, opened 2015-12-01
Added more information when failing to allocate simulator

marekcirkos:
When we fail to allocate simulator it will be way easier to understand what when wrong if we got information about runtime and type we tried to allocate.


lawrencelomax:
lgtm.

PR #84, opened 2015-12-01
Add timeout to launching and installing applications on SimDevice

mmmulani:
We've seen these methods hang on Xcode 7.1, so let's make them timeout instead.
Because the API doesn't actually error out when anything goes wrong, we run them from a background thread and shuttle the results back to the main thread.

I also tested locally with FBSimulatorControlApplicationLaunchTests.

### DIFF
--- a/FBSimulatorControl/Session/FBSimDeviceWrapper.m
+++ b/FBSimulatorControl/Session/FBSimDeviceWrapper.m
@@ -1,0 +1,101 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "FBSimDeviceWrapper.h"
+
+#import <CoreSimulator/SimDevice.h>
+
+#import "FBSimulatorError.h"
+
+const long kFBSimDeviceCommandTimeout = 30;
+
+@interface FBSimDeviceWrapper ()
+
+@property SimDevice *device;
+
+@end
+
+@implementation FBSimDeviceWrapper
+
+- (instancetype)initWithSimDevice:(SimDevice *)device
+{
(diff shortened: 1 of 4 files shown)


FBSimulatorControl/Session/FBSimDeviceWrapper.m
```diff
@@ -0,0 +1,120 @@
...
+  _device = device;
+  _lastCallID = 0;
+
+  return self;
+}
+
+- (void)scheduleInvocationInBackground:(NSInvocation *)invocation
+{
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [NSThread detachNewThreadSelector:@selector(runInvocation:) toTarget:self withObject:invocation];
```
ExtremeMan: Why NSThread and not GCD?

FBSimulatorControl/Session/FBSimDeviceWrapper.m
```diff
@@ -0,0 +1,120 @@
...
+{
+  NSAssert(![NSThread isMainThread], @"Should be on a background thread");
+
+  _lastCallID++;
+  NSInteger currentCallID = _lastCallID;
+  [invocation invoke];
+
+  dispatch_async(dispatch_get_main_queue(), ^{
+    if (_lastCallID == currentCallID) {
+      CFRunLoopStop(CFRunLoopGetCurrent());
```
ExtremeMan: Just use `dispatch_semaphore_*`.

FBSimulatorControl/Session/FBSimDeviceWrapper.m
mmmulani: GCD would mean creating new queues and destroying them since we expect the call can get stuck forever. With NSThread this is all I have to do to create a new thread

mmmulani:
updated it to use semaphores (which makes the code much nicer) but I'm still using background threads as it makes it easier to reason about and I don't have to worry about cleaning up background queues, etc.

lawrencelomax:
lgtm. May be an idea to have two constructors, one that specifies a timeout and another that uses the default timeout since people may want to vary this by machine.

PR #88, opened 2015-12-03
Share FBSimulator instances

lawrencelomax:
We are creating `FBSimulator` instances every time that the `allSimulators` property of `FBSimulatorPool` is called. In reality these objects are references and not values, as we want the process information to be updated in-place. This solves two issues:
1) Process Launch Info is only calculated once per Simulator
2) Notifications for Process lifecycle can be uniqued based on a single simulator instance, this means that notifications for Simulator and Simulator Subprocess launch can come from the Simulator itself.


marekcirkos:
lgtm

PR #91, opened 2015-12-07
Liberate Simulator Events from Sessions

lawrencelomax:
The history of events that occur during the lifecycle of a Simulator have previously been tied to the availability of a session:

- `FBSimulatorEventSink` defines a protocol for handling events that occur in the lifecycle of a simulator.
- `FBSimulatorSessionLifecycle` has been decomposed into discrete `FBSimulatorEventSink` implementations.
- `FBSimulatorSessionState` has been replaced with `FBSimulatorHistory` as there is no longer any coupling to a Session. All Session-specific information is stripped out of the history. The lifecycle of a session and of a simulator are orthogonal concerns. `FBSimulatorHistory` is a full value, without knowledge of the Simulator it is representing the history for. This means that Simulator history can serialized independently of the existence of a Simulator. This means that history for a Simulator can be looked at after a simulator has been deleted.
- `FBSimulatorSessionStateGenerator` has been replaced `FBSimulatorHistoryGenerator`. `FBSimulatorHistoryGenerator` is a simulator event sink so composes nicely with others. All of the Session handling logic has been removed.
- `FBSimulator(Helpers)` pulls out some of the helper functions that were cluttering the `FBSimulator` definition.
- `FBSimulatorEventRelay` converts observed events such as application termination into invocations of `FBSimulatorEventSink` methods. This was previously done in `FBSimulator. Events are also deduplicated here. This prevents racing of events being synthesized for unexpected application termination as well as direct termination.
- `FBSimulatorNotificationEventSink` is an Event Sink that synthesizes `NSNotifications`.
- `FBSimulatorResourceManager` takes on the responsibility of releasing resources previously done by `FBSimulatorSessionLifecycle`. This means that resources will be released outside of the context of a session, so resources can be managed without a session.
- `FBProcessInfo` is now an concrete type that is used regardless of whether `FBSimulatorControl` launched the application, or another process did. This means that applications launched by us get the full environment that `launchd_sim` wraps around the process.
- `FBProcessLaunchConfiguration` is just used for storing information about how to launch processes and is not use to represent the state of running processes, as `FBProcessInfo` fulfills this role. A mappping of processes-to-launch-info is maintained to allow for relaunching.
- `FBSimulatorHistoryGenerator` does not need to re-write history to update diagnostics for terminated applications. Since diagnostics are keyed per-process rather than per-binary as before, each launch of a binary has a separate set of process diagnostics. This also means that history is append-only, which makes streaming history a possibility.
- The distinction between process and simulator diagnostics is now clearer with separate dictionaries for each.
- `FBSimulatorHistory` conforms to `NSCoding`, meaning that it can be serialized to disk. This requires that all the objects in the graph that are connected to `FBSimulatorHistory` are also serializable: `FBProcessLaunchConfiguration` & `FBProcessInfo`. Knowledge about the current state of the world can be read on launch of the process utilizing `FBSimulatorControl`.
- `FBSimulatorSession` doesn't make any assumptions about the lifecycle of a simulator, it just serves as a way of locking access to a resource (Simulators). State transitions are implemented in terms of transitions within a class-cluster.


FBSimulatorControl/Model/FBProcessInfo.m
```diff
@@ -8,124 +8,89 @@
...
+  self = [super init];
+  if (!self) {
+    return nil;
   }
-  return self.processIdentifier == object.processIdentifier &&
-         [self.launchPath isEqualToString:object.launchPath] &&
-         [self.arguments isEqualToArray:object.arguments] &&
-         [self.environment isEqualToDictionary:object.environment];
+
+  _processIdentifier = [coder decodeInt32ForKey:NSStringFromSelector(@selector(processIdentifier))];
```
marekcirkos: Not sure if this is best approach. You will have no backward capability if you rename your selector. Not sure if it will even happen with your use case, but keep it in mind

marekcirkos:
Obviously I don't follow flow changes so lets hope for the best :)

FBSimulatorControl/Model/FBProcessInfo.m
lawrencelomax: I'm not too worried about migration right now, so even if the 'schema' changes and the decoding fails we can just start again.

lawrencelomax:
giphy

Looks like #98 solved much of the flakiness we were seeing in the tests!

PR #96, opened 2015-12-09
Improve App Launch Fixtures & Error Handling

lawrencelomax:
In #95 I'm having difficulty diagnosing why some of the App Launches are failing, this should help diagnose the callsite that is likely failing.


lawrencelomax:
My suspicion that the the code fetching information about the binary architectures is flaky looks to be correct based on the last Travis run.

lawrencelomax:
Creating all of the System Applications in one giant concurrent `map` is probably a bad idea, especially since the arch fetching will require many subprocesses. Creating the `FBSimulatorApplication` instances on demand seems to be the simplest approach since I can't see many circumstances where clients will need to know the archs of all the System applications. I did think about lazy-loading the archs, but this adds more complexity and gets in the way of the immutability of `FBSimulatorApplication`.

marekcirkos:
lgtm

PR #97, opened 2015-12-10
Marking FBSimulatorControl simulators

marekcirkos:
Marking FBSimulatorControl simulators with FBSIMULATORCONTROL_SIM_UDID env. It will make it easer to FBSimulatorControl with other simulator managers.


lawrencelomax:
👌

PR #98, opened 2015-12-11
Read Mach-O Header

lawrencelomax:
I believe that the `FBTask` implementation is a bit wonky on Travis which suggests some raciness that needs further investigation. However this 1) Doesn't repro for me locally 2) Only appears a bunch on the arch fetching. The arch fetching can be [re-implemented without too much effort](http://lowlevelbits.org/parse-mach-o-files/) by reading the header directly.


marekcirkos:
lgtm

PR #99, opened 2015-12-11
Optionally persist Simulator history to file

lawrencelomax:
This can be used to re-inflate the knowledge of the history of a Simulator when a Simulator is created. Very handy for the CLI use case, which can work via multiple invocations of a process linking `FBSimulatorControl`.


marekcirkos:
lgtm

marekcirkos:
Lgtm, however you should fix failure.

PR #100, opened 2015-12-11
Fix Application Process Fetch Interactions

lawrencelomax:
Re-launch is a little messed up since it relies on an Application existing, the logic was a little faulty here is this meant it wasn't always possible to fetch the last-launched app process. This fixes that.


marekcirkos:
lgtm

lawrencelomax:
It looks instances of Simulators [are accumilating over](https://travis-ci.org/facebook/FBSimulatorControl/builds/96709849#L382) tests which is causing notifications to be duplicated. There's probably a cycle in the Simulator infrastructure somewhere that is causing this.

lawrencelomax:
So it looks like most of this was due to ordering issues with notifications with changes in shutdown and notification receiving behaviour. This has been fixed with some juggling around of the assertion methods.

PR #105, opened 2015-12-16
Query History by State Change

lawrencelomax:
Sometimes you may want to query Simulator history in order to find out the last time a particular event occurred. In the case of logs you may want to know last time a Simulator was booted in order to narrow down information and discard pre-last-boot information. By adding queries by last-change-to-state it's possible to find this information.

In addition, the recursive description has been improved so that the ordering is correct and less-noisy descriptions are used.


marekcirkos:
lgtm

PR #108, opened 2015-12-17
Use directory of asl files

lawrencelomax:
Turns out that there's not any need to use a specific asl file since `asl` is just fine with a whole directory. It looks like different files are used at different times anyway and this could mean that certain messages go to different paths. Also made sure that all logs are forwarded with the debug log level.


marekcirkos:
lgtm

PR #109, opened 2015-12-18
Weak Link Private Frameworks & Load at runtime

lawrencelomax:
The hard-coding of the `rpath` in the `FBSimulatorControl.framework` binary has been a big problem in terms of flexibility, but it looks like we won't need it anymore :+1: 

I did consider prototype this using the [method that `xctool` does in it's `main.m`](https://github.com/facebook/xctool/blob/master/xctool/xctool/main.m#L37). This did work and in some senses is more flexible since the paths to the Private frameworks are searched for instead of provided as relative paths to the developer directory. However, since `FBSimulatorControl` is commonly loaded as a framework, I thought it would be less invasive to load the Frameworks at runtime, rather than calling `exec`. 

The only major downside to loading the Frameworks directly, is that the symbols for classes are not resolved and `NSClassFromString` must be used to obtain the relevant classes at runtime. This isn't too much of an issue since very few class methods/constructors for weak Frameworks are actually called and I doubt many consumers of `FBSimulatorControl` will ever need to access `CoreSimulator` and friends directly. Calling instance methods is just fine since selectors are loaded.

- [x] Remove all of the Developer directories from `rpath`.
- [x] Remove Private Frameworks from 'Link Binary with Libraries'
- [x] Use `-weak_framework DVTFoundation -weak_framework DVTiPhoneSimulatorRemoteClient -weak_framework CoreSimulator`.
- [x] Use `__attribute__((constructor))` to hook a Framework loading function.
- [x] Load Frameworks
- [x] Replace Direct Class symbol access with `NSClassFromString`
- [ ] Update Documentation.


marekcirkos:
Nice work!

PR #110, opened 2015-12-18
Additional Error Messaging when failing to kill processes.

lawrencelomax:
Error messaging on termination can be somewhat cryptic, this should fix that.


marekcirkos:
lgtm

PR #111, opened 2015-12-18
Explicit Matching of Device Families

lawrencelomax:
Given the advent of `tvOS` things can end up in a sorry state if an iPhone device type is paired with a `tvOS` runtime. This change will improve the knowledge that exists in each configuration variant so that runtimes are matched to their appropriate device types

- [x] Match Runtimes against Device Families in Device Type.
- [x] Provide error messages to callers who obtain runtimes and device types.
- [x] Add 'Device Family' knowledge to Configuration Variants.
- [x] Add Recent iOS versions.
- [x] Add tvOS device families.
- [x] Add watchOS device families.
- [x] Pick configuration defaults on runtime availability.
- [x] Provide inverse mapping of runtime to configuration.
- [x] Allocation precondition for device/runtime availability.
- [x] Failable Arguments for Device Scale.
- [x] Tests for watch and tv simulator types.


marekcirkos:
lgtm

PR #118, opened 2015-12-21
Add Log Levels with additional error & debug logging

lawrencelomax:
Adding log levels will allow for finer grained logging in the CLI use-case. Additionally, the addition of some logs will make it far easier to track down what the causes of failures are during a run of `FBSimulatorControl`.


marekcirkos:
lgtm

PR #119, opened 2015-12-21
Break out fbsimctl core to FBSimulatorControlKit for testing

lawrencelomax:
`FBSimulatorControlKit` is the backbone of `fbsimctl`. Breaking out this into a Framework means that code on which `fbsimctl` depends can be Tested directly.

Unfortunately, I can't find a good way of integrating `FBSimulatorControlKit` as a dependency of `fbsimctl` without linker issues. For the time being, `FBSimulatorControlKit` acts as a container for the purposes of producing a test target. The source files are compiled for `fbsimctl` as well which means that `FBSimulatorControlKit` isn't a dependency of `fbsimctl`, but it's sources are shared.


marekcirkos:
lgtm

PR #121, opened 2015-12-22
Query Combinations

lawrencelomax:
Queries can be more expressive by performing a union along queries of the same case and an intersection across cases:

- `"iPhone 5" "iPhone 6"` will find both iPhone 5 and iPhone 6 Simulators.
- `"iPhone 5" "iPhone 6" B8EEA6C4-841B-47E5-92DE-014E0ECD8139` will find iPhone 5 and 6 Simulators that match the given UDID.
- `booting booted "iPad 2"` will find all booting or booted iPad 2 Simulators.


marekcirkos:
lgtm

PR #122, opened 2015-12-22
Added wait for required processes when booting simulator

marekcirkos:


FBSimulatorControl/Interactions/FBSimulatorInteraction.m
```diff
@@ -91,6 +91,26 @@ - (instancetype)bootSimulator
...
+    @[
+      @"SpringBoard",
+      @"com.apple.accessibility.AccessibilityUIServer",
+      @"com.apple.audio.SystemSoundServer-iOS-Simulator",
+      @"AssetCacheLocatorService",
+      @"MobileCal",
+      @"medialibraryd",
+    ]];
+    BOOL didStartAllRequiredProcesses = [[NSRunLoop mainRunLoop] spinRunLoopWithTimeout:60 untilTrue:^BOOL{
+      NSSet *runningProcesses = [NSSet setWithArray:[simulator.processQuery processesWithLaunchPathSubstring:nil]];
```
lawrencelomax: You don't have to get all processes there is `subprocessesOf:` query that you can use on the `launchd_sim` process identifier that you have from the `FBSimulatorLaunchInfo`.

In fact you could end up querying for  (and finding) these subprocesses of a *different* simulator.

FBSimulatorControl/Processes/FBProcessQuery.m
```diff
@@ -223,7 +223,7 @@ - (NSArray *)processesWithLaunchPathSubstring:(NSString *)substring
     if (!info) {
       return YES;
     }
-    if ([info.launchPath rangeOfString:substring].location == NSNotFound) {
+    if (substring && [info.launchPath rangeOfString:substring].location == NSNotFound) {
```
lawrencelomax: I'm happy to do this, but you should update the header documentation `@param substring the substring that must exist in the launch path.` to include that omiting the param will return all processes.

lawrencelomax:
Looks like there's a build error, you should use `failBool` instead of `fail` since interactions always return a `BOOL`

FBSimulatorControl/Interactions/FBSimulatorInteraction.m
marekcirkos: Ups :)

marekcirkos:
Is this test `[FBSimulatorLaunchInfoTests testUpdatesTerminationInformation]` flaky? I can't reproduce it

lawrencelomax:
A bit more context to the intent of this PR:
Just because a device is in a `booted` state, does not mean that the the device is truly ready to have applications launched and the Simulator otherwise messed around with. It's entirely possible to boo daemons before `Springboard` has fully launched, however it's probably best to ensure that `Springboard` is ready (along with a few other Services) before attempting to interact with the Simulator.

Given that the `launchd_sim` process identifier is known during the `bootSimulator` method, the `FBSimulatorLaunchInfo` object can be used to find the appropriate `launchd_sim` subprocesses and put this inside a wait-loop.

lawrencelomax:
With regards to the flakyness, I think it could be mainly due to the long run-time of Travis tests relative to the performance characteristics that you'll see elsewhere. 60 seconds is probably a bit too low when the average test is taking around 70s.

PR #123, opened 2015-12-22
Polyfill for -[SimDevice addVideo:error:]

lawrencelomax:
Since `addVideo:error:` was added to `SimDevice` in recent versions of Xcode, we can use it directly instead of relying on the `Shimulator` hack. Instead of removing this hack entirely, we can call out to it in the event that the API doesn't exist at runtime.


marekcirkos:
lgtm

PR #124, opened 2015-12-22
Opt-In for autoload of FBSimulatorControl Private Frameworks.

lawrencelomax:
It's possible that the constructors are re-entrant and this causes issues with `otest-query-osx` used in `xctool`. It's better to place the onus of class loading on the consumer and additionally make it a precondition of `FBSimulatorContol`.

Additionally this removes the `+[DVTPlatform loadAllPlatformsReturningError:]` call. This call was actually *failing* but was being masked by the fact that the precondition method was erroneously never evaluated.


marekcirkos:
lgtm

PR #135, opened 2015-12-29
Remove Simulator Application from FBSimulatorControlConfiguration

lawrencelomax:
This is effectively a Global requirement since it is derived from the `xcode-select` path. For this reason it makes more sense to be a global and not bloat the Configuration object.


marekcirkos:
lgtm

PR #136, opened 2015-12-29
Add Debug Logging after creation

lawrencelomax:
Passing the Debug logging flags after the creation of the global logger object is inconvenient since it makes controlling this at runtime impossible without making assumptions about initialization order in the CLI. `asl` makes it easy to pipe to additional files.


marekcirkos:
lgtm

PR #137, opened 2015-12-29
Update Global Configuration

lawrencelomax:
Since the `+[NSObject initialize]` method may be called during the parsing of arguments, it is currently essential to have the debug logging flag enabled in order for the information be printed to `stderr`. `setenv` works well for the tests and will also work fine from the CLI.


marekcirkos:
lgtm

PR #138, opened 2015-12-29
CLI Flag for enabling debug logging

lawrencelomax:
Since calling `FBSimulatorControl` classes can invoke the initializer and therefore load the Frameworks, this can lead to the omission of Framework loading debug messages, making sure that the Debug flag is known about up-front will prevent Parsing having to worry about ordering.


marekcirkos:
lgtm

PR #139, opened 2015-12-29
Continuous console output for runners

lawrencelomax:
Since a `Query` can result in multiple Simulators being matched it's important to be able to output success messages incrementally (for example in Booting). It's also important to be able to distinguish between 'logger' messages and 'output' messages; logger messages are informational and fine grained, output messages represent the result of an API call which would closely map to the return value of API calls.


marekcirkos:
lgtm

PR #143, opened 2015-12-31
Added wait with timeout when looking for launchInfo of FBSimulatorControled simulator

marekcirkos:
It's good to have the state update for simulators that are not launched directly via FBSimulatorControl, but FBSimulatorControl is used as a helper to manage them and for some helpful APIs.


lawrencelomax:
lgtm

PR #145, opened 2016-01-04
Move Query & Format to Top Level of an Action

lawrencelomax:
`Action` has moved to a struct with members `Query`, `Format` & `Interaction`. This fulfills a few needs:
- There is far less duplication of performing queries on an `Action` since all actions have a query.
- The output format of all actions uses the `Format`
- Future `Interaction` cases that don't rely on any associated types are simpler to implement
- `Interaction` cases map closer to `FBSimulatorInteraction`.
- `Help` should be easier to define.


marekcirkos:
lgtm

PR #146, opened 2016-01-04
Non-Arbitrary Ordering for Process Launch Configs in History

lawrencelomax:
The `allProcessLaunches` value previously relied on the ordering of the `allValues` array in a dictionary, which has arbitrary ordering. Instead the array of `allProcessLaunches` can be constructed from the ordered `processLaunchConfigurations` collection.


marekcirkos:
lgtm

PR #148, opened 2016-01-05
Create FBSimulatorLaunchConfiguration

lawrencelomax:
Adds `FBSimulatorLaunchConfiguration` which is an immutable configuration object passed to `bootSimulator:`. This achieves a number of things:
- `FBSimulatorConfiguration` was never the right place to have `scale` and `locale` arguments, is configuration for the pre-launch interactions of a Simulator
- A common container for options to `bootSimulator` as I plan to add more configurability to the boot process.
- Removal of configuration of a simulator from `allocateSimulator`. The `FBSimulatorInteraction+Setup` suite of interactions should be used directly prior to `bootSimulator` instead. Allocation does not imply pre-boot configuration and this is unnecessarily surprising.
- A neat place to factory arguments to the Simulator Application process.

### DIFF
--- a/FBSimulatorControl/Configuration/FBSimulatorConfiguration.m
+++ b/FBSimulatorControl/Configuration/FBSimulatorConfiguration.m
@@ -540,7 +540,7 @@
 
 - (NSUInteger)hash
 {
-  return self.deviceName.hash | self.osVersionString.hash;
+  return self.deviceName.hash ^ self.osVersionString.hash;
 }
 
 - (BOOL)isEqual:(FBSimulatorConfiguration *)object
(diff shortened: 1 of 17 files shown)


FBSimulatorControl/Configuration/FBSimulatorConfiguration.m
```diff
@@ -584,16 +536,11 @@ - (NSString *)osVersionString
...
-{
-  return self.scale.scaleString;
-}
-
 #pragma mark NSObject
 
 - (NSUInteger)hash
 {
-  return self.deviceName.hash | self.osVersionString.hash | self.locale.hash | self.scaleString.hash;
+  return self.deviceName.hash | self.osVersionString.hash;
```
mmmulani: should be XOR I think

PR #150, opened 2016-01-05
Boot Simulator Directly and connect to Framebuffer

lawrencelomax:
`SimulatorKit` has a very interesting opportunity for `FBSimulatorControl`. The primary motivation for using `Simulator.app` instead of booting directly with `-[SimDevice bootWithOptions:error]` was that there is a framebuffer connected when `Simulator.app` launches. 

Using a similar methodology to create a `SimDeviceFramebufferService` as Xcode Playgrounds does, it's possible to create a Framebuffer, boot the simulator, then listen for events on the framebuffer. These events are `CGImageRef`s so can be recorded to images and video.

- [x] `FBSimulatorLaunchConfiguration` control
- [x] Basic implementation in `bootSimulator`
- [x] Lots and lots of code-level documentation for how and why this works.
- [x] Updates to the README to describe the differences between 'direct' and 'container' launches.
- [x] Event Sink events for framebuffer lifecycle.
- [x] Add framebuffer to the simulator instance.
- [x] Debug `NSWindow` canvas for live framebuffer view.
- [x] Video Framebuffer recorder.
- [x] Use Scale variable on Framebuffer to encode at scaled resolution.

PR #153, opened 2016-01-07
Killing Simulators accounts for Simulator.app-less booted Simulators.

lawrencelomax:
Given that it will soon be possible to boot Simulators (and get a framebuffer #150) without a `Simulator.app` process, it's important that  `FBSimulatorTerminationStrategy` can account for this. 

This change improves things in this regard:
1) Makes `FBSimulatorTerminationStrategy` a bit simpler and easier to comprehend
2) Documentation
3) Doesn't couple Application Termination to the `Simulator.app`, processes that are applications can get use `NSWorkspace` termination for free.
4) Fallback behaviour for `NSWorkspace` termination.
5) Separating the concerns of 'killing the Simulator.app' process vs. 'shutting down a Simulator'. This will be important in #150 as getting `CoreSimulator` into a stable state on launch may not require killing `Simulator.app` processes.


marekcirkos:
lgtm

PR #155, opened 2016-01-07
Separate concerns of Simulator Container & Booted Simulator

lawrencelomax:
After #150 we won't be able to assume that a 'launched' Simulator is one that has a `Simulator.app` process as well as all of the associated information in `CoreSimulator` as well as the `launchd_sim` process. 

This means that the concept of a 'Container' application needs to be introduced. Currently `Simulator.app` is the only kind of container (in the future we could make a container and indeed `fbsimctl` could also be one). In addition the criteria for the 'bootedness' of a Simulator should change according to the manner in which the Simulator was booted.

Also:
- [x] Launch of Container & of `launchd_sim` have the process info on their respective event sink methods.
- [x] Removal of `FBSimulatorLaunchInfo` since it would otherwise be a factory for the `launchd_sim` process info.
- [x] Updating of the Simulator launch info based on `SimDevice` notifications.
- [x] Improvements to the tests to account for the different kinds of booting.


FBSimulatorControl/Processes/FBProcessQuery+Helpers.h
```diff
@@ -10,6 +10,7 @@
 #import <FBSimulatorControl/FBProcessQuery.h>
 
 @class FBProcessInfo;
+@class FBProcessInfo;
```
mmmulani: duplicate

FBSimulatorControl/Events/FBSimulatorEventSink.h
```diff
@@ -11,9 +11,9 @@
 
 @class FBAgentLaunchConfiguration;
 @class FBApplicationLaunchConfiguration;
+@class FBProcessInfo;
```
mmmulani: duplicate again?

mmmulani:
lgtm though!

PR #158, opened 2016-01-13
Use NSWorkspace API for Launching Simulator Applications

lawrencelomax:
`NSWorkspace` provides a very nice API for launching Applications, and will even fit our requirements of:
- Providing launch arguments
- Providing launch environment
- Always launching a new Application instance, even if one or more is already running.

Weaning ourselves of `FBTask` is a good idea in general since It's another thing to go wrong™️. We'll get better error messaging from the `NSWorkspace` API. Additionally, the work for `fbsimctl` to ensure that the `Simulator.app` process lifetime is not dictated by the lifetime of the containing `FBSimulatorControl` process means that `FBSimulatorTerminationStrategy` is the most suitable place for knowledge for how to terminate an application, rather than a subprocess.


lawrencelomax:
Well this is sad, looks like we get a useless unknown error on travis, can't repro this locally.

lawrencelomax:
Adding as an experimental feature.

PR #163, opened 2016-01-14
Use --state= prefix for state queries

lawrencelomax:
In order to make parsing context-free and to remove any ambiguity about what `shutdown` means, I'm changing `shutdown` the query to `--state=shutdown` so that it isn't confused with `shutdown` the interaction.


asm89:
lgtm

PR #166, opened 2016-01-14
Use exit(3) for a return code on success or failure

lawrencelomax:
Failing commands should result in a return code.


phyllipecesar:
LGTM, we can have specific return codes later.

PR #168, opened 2016-01-18
Extract underlying error when executing FBSimDeviceWrapper methods

lawrencelomax:
In the `TimeoutResiliance` case for `FBSimDeviceWrapper` we don't do anything with the error from `SimDevice` so the underlying error cause can get missed, this is problematic on Xcode 7.2 which still seems to have issues with `installApplication:`

### DIFF
--- a/FBSimulatorControl/Utility/FBSimDeviceWrapper.m
+++ b/FBSimulatorControl/Utility/FBSimDeviceWrapper.m
@@ -87,22 +87,23 @@
 {
   NSAssert([NSThread isMainThread], @"Must be called from the main thread.");
 
-  NSError *__autoreleasing innerError = nil;
-  NSError *__autoreleasing *innerErrorPointer = &innerError;
+  NSError *innerError = nil;
   NSInvocation *invocation = [NSInvocation invocationWithMethodSignature:[self methodSignatureForSelector:@selector(installApplication:withOptions:error:)]];
   [invocation setTarget:self.simulator.device];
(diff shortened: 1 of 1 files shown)


FBSimulatorControl/Utility/FBSimDeviceWrapper.m
```diff
@@ -87,22 +88,23 @@ - (BOOL)installApplication:(NSURL *)appURL withOptions:(NSDictionary *)options e
...
-  NSError *__autoreleasing *innerErrorPointer = &innerError;
+  NSError *innerError = nil;
   NSInvocation *invocation = [NSInvocation invocationWithMethodSignature:[self methodSignatureForSelector:@selector(installApplication:withOptions:error:)]];
   [invocation setTarget:self.simulator.device];
   [invocation setSelector:@selector(installApplication:withOptions:error:)];
   [invocation setArgument:&appURL atIndex:2];
   [invocation setArgument:&options atIndex:3];
-  [invocation setArgument:&innerErrorPointer atIndex:4];
-  error = innerErrorPointer;
+  [invocation setArgument:&innerError atIndex:4];
```
marekcirkos: Did you test it? It was crashing before that change with innerErrorPointer was made.

lawrencelomax: Tested by faking a broken install path & launch path. The problem is that `__autoreleasing` which meant that the lifetime of the `NSError` was reduced.

lawrencelomax:
@facebook-github-bot shipit

PR #169, opened 2016-01-18
Full JSON Output CLI

lawrencelomax:
- [x] Replace `Writer` on the `Runner` side with a higher-level representation, `EventReporter` that can serialize to a writer for an Event that occurs.
- [x] Add Event Names
- [x] Human Readable Reporter
- [x] JSON Reporter
- [x] Move JSON to `Format` rather than global configuration.

### DIFF
--- a/fbsimctl/FBSimulatorControlKit/Sources/EventReporter.swift
+++ b/fbsimctl/FBSimulatorControlKit/Sources/EventReporter.swift
@@ -1,0 +1,107 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+import Foundation
+import FBSimulatorControl
+
+public typealias EventReporterSubject = protocol<FBJSONSerializationDescribeable, FBDebugDescribeable>
+
+public enum EventName : String {
+  case Boot
+  case Shutdown
+  case Diagnostic
+  case Delete
+  case Install
+  case Launch
+  case Approve
+  case StateChange
+}
+
+public enum EventType : String {
+  case Started = "Started"
+  case Ended = "Ended"
+  case Discrete = ""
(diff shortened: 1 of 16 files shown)


fbsimctl/FBSimulatorControlKit/Sources/EventReporter.swift
```diff
@@ -0,0 +1,214 @@
...
+        case .State:
+          return simulator.stateString
+        case .ProcessIdentifier:
+          guard let process = simulator.launchdSimProcess else {
+            return "no-process"
+          }
+          return process.processIdentifier.description
+        }
+      }
+      return tokens.joinWithSeparator(" ")
```
phyllipecesar: What does it mean? a name "a b" could be interpreted as two tokens? Should we use commas?

fbsimctl/FBSimulatorControlKit/Sources/EventReporter.swift
```diff
@@ -0,0 +1,214 @@
...
+    do {
+      self.writer.write(
+        try self.json.serializeToString([
+          "simulator" : self.simulator.jsonSerializableRepresentation(),
+          "event_name" : eventName.rawValue,
+          "event_type" : eventType.rawValue,
+          "subject" : subject.jsonSerializableRepresentation()
+        ])
+      )
+    } catch {
```
phyllipecesar: Sketchy =( Can we log anything?

fbsimctl/FBSimulatorControlKit/Sources/EventReporter.swift
lawrencelomax: That's a good point, I think the simplest way for now is to quote tokens that have spaces in them in the same way that we separate argument tokens i.e. `'iPhone 5' list`

fbsimctl/FBSimulatorControlKit/Sources/EventReporter.swift
lawrencelomax: I was thinking we `abort` instead, it's programmer error if an event is passed that is un-seralizable.

lawrencelomax:
@facebook-github-bot shipit

PR #170, opened 2016-01-18
FBApplicationLaunchConfiguration defined in terms of Bundle ID

lawrencelomax:
It should be possible to launch an Application without the location of the source Application. The path to the original binary makes less sense considering the Application is installed relative to the `data` directory of Simulator, so the original location of the `Application` binary isn't used when the Application is a launched on the Simulator.


FBSimulatorControl/Configuration/FBProcessLaunchConfiguration.m
```diff
@@ -195,19 +200,20 @@ - (void)encodeWithCoder:(NSCoder *)coder
...
-  [coder encodeObject:self.application forKey:NSStringFromSelector(@selector(application))];
+  [coder encodeObject:self.bundleID forKey:NSStringFromSelector(@selector(bundleID))];
 }
 
 #pragma mark NSObject
 
 - (NSUInteger)hash
 {
-  return [super hash] | self.application.hash;
+  return [super hash] | self.bundleID.hash;
```
mmmulani: should be an xor

lawrencelomax:
@facebook-github-bot shipit

PR #172, opened 2016-01-19
Use FBSimulatorLaunchConfiguration to configure simulator before boot

lawrencelomax:
This allows for the locale to be set before booting as well as the scale for the boot itself. This will also be able to account for options in the future when I finally get round to finishing #150


lawrencelomax:
@facebook-github-bot shipit

PR #174, opened 2016-01-20
Use propertiesOfApplication instead of installedApps

lawrencelomax:
Saves pulling down a massive dictionary.


lawrencelomax:
@facebook-github-bot shipit

PR #175, opened 2016-01-20
Improve launchd_sim subprocess killing

lawrencelomax:
Instead of just having `killApplication:` to kill applications, we should expose an API to do this for any subprocess of the simulator. We can add additional checks so that we don't kill random processes on a host and also use the existing `FBSimulatorTerminationStrategy`

Additionally, it looks like there is a race condition in the relaunch case:
1) An app is launched and has a pid
2) The app is terminated
3) Wait for the pid to disappear
4) Relaunch the app with the same launch configuration
5) The relaunched app will have the same pid as 1)
6) We try and fetch the process info for the pid, but it has died.

I can only assume that `CoreSimulator` will talk to the `launchd_sim` inside the Simulator and as the process was killed externally, it has to update it's knowledge of the world to determine that the process has died.

### DIFF
--- a/FBSimulatorControl/Interactions/FBSimulatorInteraction+Applications.h
+++ b/FBSimulatorControl/Interactions/FBSimulatorInteraction+Applications.h
@@ -30,22 +30,6 @@
 - (instancetype)launchApplication:(FBApplicationLaunchConfiguration *)appLaunch;
 
 /**
- Unix Signals the Application.
-
- @param signal the unix signo to send.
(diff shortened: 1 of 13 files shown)


lawrencelomax:
Fixed by waiting on `launchctl` in #176, tests now pass.

mmmulani:
lgtm

lawrencelomax:
@facebook-github-bot shipit

PR #176, opened 2016-01-21
Add Wrapper to Simulator's launchctl

lawrencelomax:
In #175 I discovered a race that can occur between the killing of a process and the Simulator's knowledge of this process in `launchctl`.

By looking at `launchctl`s knowledge of the world by spawning a `launchctl` process from `FBSimulatorControl` (via `CoreSimulator`) we can know whether `launchctl` is aware of the process or not. I am looking to add features to this class over time as there's plenty of useful information in `launchctl` that we currently don't have access to in `FBSimulatorControl`.


lawrencelomax:
@facebook-github-bot shipit

PR #177, opened 2016-01-21
Add method to return home directory for last app launched

nqmtuan:
This pull request is to add a method for FBSimulatorControl to return the home directory for the last app launched.


nqmtuan:
@facebook-github-bot shipit

PR #181, opened 2016-01-23
Fixing error string

michaelpetrov:
Fixing error string to correctly report both the actual and expected state of the simulator


lawrencelomax:
Nice Catch!

lawrencelomax:
@facebook-github-bot shipit

PR #186, opened 2016-01-28
Utility method for updating a plist in place

lawrencelomax:
There's a bunch of methods that have a common way of updating/creating a plist, so this should be pulled out into a helper method.

@mmmulani you may want to check the `FBLaunchWatchdogExceptions` value, it looks like the old implementation will just overrwrite the value with the last item in the array, I think this fixes that.


mmmulani:
lgtm!!

lawrencelomax:
@facebook-github-bot shipit

PR #187, opened 2016-01-28
Remove FBSimulatorSession

lawrencelomax:
This is a class with no purpose that adds confusion to the API. `FBSimulator` has all of the interesting APIs and the Session has just hung around since the early days. There's no use for it any more so it's well worth reducing the API overhead.


lawrencelomax:
@facebook-github-bot import

PR #189, opened 2016-01-29
Remove FBSimulatorWindowTiler

lawrencelomax:
See #188. The tiler was only ever needed to place Simulator windows so they didn't overlap when recording video. Using `FBFramebufferVideo` will avoid the need to do this in the first place.


lawrencelomax:
@facebook-github-bot shipit

PR #190, opened 2016-01-29
Create and Register a 'IndigoHIDRegistrationPort' Port

lawrencelomax:
This is another missing piece of the Framebuffer implementation, meaning that the implementation is similar to the startup path that `Simulator.app` takes.

Before this change, it was possible that `[UIATarget tap]` (used by WebDriverAgent) would not work when a Simulator is booted in this manner. The actual event dispatch to the HID Subsystem occurs at `IOHIDEventSystemClientDispatchEvent`, and will fail without a 'Indigo HID system' registered with `backboardd`.


lawrencelomax:
@facebook-github-bot import

PR #191, opened 2016-01-29
Improve FBFramebufferVideo robustness

lawrencelomax:
I've been having difficulty with the reliability of the writing of video files when `-[AVAssetWriter movieFragmentInterval]` is set. The the following error will occur in `-[AVAssetWriterInputPixelBufferAdaptor appendPixelBuffer: withPresentationTime:]`

```Jan 29 18:19:59  FBSimulatorControl[61710] <Error>: Failed to append frame at time 16.066311 seconds of pixel buffer with error Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSUnderlyingError=0x1095aa460 {Error Domain=NSOSStatusErrorDomain Code=-16341 "(null)"}, NSLocalizedFailureReason=An unknown error occurred (-16341), NSLocalizedDescription=The operation could not be completed}
Jan 29 18:19:59  FBSimulatorControl[61710] <Error>: Failed to append frame at time 16.082559 seconds of pixel buffer with error Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSUnderlyingError=0x1095aa460 {Error Domain=NSOSStatusErrorDomain Code=-16341 "(null)"}, NSLocalizedFailureReason=An unknown error occurred (-16341), NSLocalizedDescription=The operation could not be completed}```

Once one error occurs, it'll keep happening for the rest of the recording session. I can't repro this when the `movieFragmentInterval` isn't set. By setting the interval to 5 seconds, it'll repro pretty consistently. You can switch on video recording in tests by launching them with the environment variable `FBSIMULATORCONTROL_RECORD_VIDEO` set to `YES`.

I've been unable to repro this with the interval set to 10s, even in `-[FBSimulatorLaunchTests testLaunchesMultipleSimulators]`, which should be particularly taxing since it launches three simulators and encodes them concurrently on separate queues. Perhaps setting the interval too low means that there isn't enough time to finish the write between intervals. I'll also create a radar to get better error reporting since `An unknown error occurred (-16341)` isn't very helpful.


lawrencelomax:
Had a great deal of help with diagnosing the underlying cause of why setting a `movieFragmentInterval` could result in increased failure to write with @rfistman. Much of this comes down to the ordering and values of the `CMTime` passed to `-[AVAssetWriterInputPixelBufferAdaptor appendPixelBuffer:withPresentationTime:]`. I've documented some of @rfistman's advice in the implementation.

If media timing still proves to be troublesome then we can probably add some assertions or handling of inputs to ensure all timing is sequential.

lawrencelomax:
@facebook-github-bot import

lawrencelomax:
Merging for now, can follow up with further improvements in another PR.

PR #192, opened 2016-02-01
Use appropriate option set for FBSimulatorAllocationOptionsEraseOnFree

lawrencelomax:
This was previously using the wrong option set.


lawrencelomax:
@facebook-github-bot shipit

PR #194, opened 2016-02-02
Add Configuration to FBProcessTerminationStrategy

lawrencelomax:
`FBProcessTerminationStrategy` has a bunch of behaviours on top of `kill` making these explicit with an option set means that behaviour is more understandable. `FBSimulatorManagementOptionsUseProcessKilling` no longer makes sense as it is always the fallback.

Additionally, the relaunch interaction could result in strange behaviour when the failure to kill an existing application was suppressed. As a result, the app wouldn't be re-launched at all.


lawrencelomax:
@facebook-github-bot shipit

PR #195, opened 2016-02-02
Repush the last video frame at Current Time

lawrencelomax:
When flushing the video's buffer, it can look like the video is cut short if the simulator doesn't redraw for some time. When tearing down the Video writer, we now re-push the last frame with the current time to make this idleness obvious in the video.


lawrencelomax:
@facebook-github-bot shipit

PR #196, opened 2016-02-02
Rename Logs to Diagnostics

lawrencelomax:
Diagnostics is much more descriptive since it covers images, videos and other non-textual output. It can also be ambiguous relative to `FBSimulatorLogger`


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
reviewed IRL

PR #197, opened 2016-02-03
Concurrent & Prefixed Logging

lawrencelomax:
Given that we have background queues that will want to log, we should be able to use ASL to log with separate clients on these queues. Also added some prefixing and additional video logging.


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
Reviewed IRL by @mmmulani

PR #199, opened 2016-02-03
Framebuffer Completion Wait

lawrencelomax:
It can be frustrating when the teardown of the `FBSimualtorControl` process races with the the final writing of a video to disk, as this can lead to incomplete videos.

By using `dispatch_group`s for the semaphore-like semantics and `dispatch_wait` we can pass down a group to `FBSimulatorFramebufferDelegate`'s so that they perform some teardown work, asynchronous with the main thread.

### DIFF
--- a/FBSimulatorControl/Framebuffer/FBFramebufferVideo.m
+++ b/FBSimulatorControl/Framebuffer/FBFramebufferVideo.m
@@ -82,7 +82,6 @@ static const Float64 FBFramebufferFragmentIntervalSeconds = 5;
 @property (nonatomic, strong, readwrite) FBFramebufferVideoItem *lastItem;
 
 @property (nonatomic, strong, readwrite) AVAssetWriter *writer;
-@property (nonatomic, strong, readwrite) AVAssetWriterInput *input;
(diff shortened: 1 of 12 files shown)


lawrencelomax:
@facebook-github-bot import

FBSimulatorControl/Framebuffer/FBFramebufferVideo.m
```diff
@@ -340,23 +322,20 @@ - (BOOL)createAssetWriterAtPath:(NSString *)videoPath size:(CGSize)size startTim
   return YES;
 }
 
-- (void)teardownWriter
+- (void)teardownWriterWithGroup:(dispatch_group_t)teardownGroup
 {
-  AVAssetWriter *writer = self.writer;
-  AVAssetWriterInput *input = self.input;
-  NSURL *videoFile = writer.outputURL;
-  self.writer = nil;
```
lawrencelomax: > *"I'm certain that the group should be retained"*

@jspahrsummers *Thursday, 4th February, 2016*

jspahrsummers: hwæt

PR #203, opened 2016-02-08
Add a HTTP Server to fbsimctl

lawrencelomax:
Sockets are awesome, but HTTP is the lingua-franca of getting stuff done with networked machines. We can re-use much of the existing infra that we have in `fbsimctl` in terms of the Request/Response lifecycle and add a 'HTTP' listener in addition to stdio and sockets.

Using carthage instead of cocoapods for fetching and building external dependencies is preferable is it won't enforce it's view of how to construct `fbsimctl` project files. It could also help with getting a non-binary homebrew formula.

- [x] Rename `interact` to `listen`
- [x] Add `--http` as an alternative to `--socket` for listening.
- [x] Add [a suitable HTTP library](https://github.com/glock45/swifter)
- [x] Create a new `Relay` for HTTP.
- [x] Re-distributions of licenses for dependencies.
- [x] Add simple API for extracting values from `JSON`
- [x] Pull in dependencies for `fbsimctl` ci build in travis.
- [x] Implement/Generate endpoints and parsers for interactions.


lawrencelomax:
Since I'm not re-distributing the source or binaries of 3rd party deps, there's nothing hairy about licensing.

lawrencelomax:
`fbsimctl --state=booted listen --http 8090` then `curl -X "POST" "http://localhost:8090/relaunch" \
	-H "Content-Type: application/json" \
	-d "{\"bundle_id\":\"com.apple.mobilesafari\",\"bundle_name\":\"Safari\",\"arguments\":[],\"environment\":{}}"` 
lgtm on a booted simulator.

lawrencelomax:
Reviewed by @mmmulani IRL

PR #204, opened 2016-02-09
Launch vs. Relaunch Interactions

lawrencelomax:
Tightens up the semantics around 'launching' and 're-launching':
- Launch implies that the interaction will fail if the app is currently launched
- Re-launch implies that the interaction will ensure the app is killed before launching it.

This means that for many use-cases will be the preferable interaction: *'I just want to start this app, or re-start it if it's running'*. Also adds the ability to 're-launch' with a different launch configuration by querying the process table for a launched application with the same Bundle ID. Install is also simpler since the install interaction won't fail if the application is a System Application.

Additionally, the tests for launch and re-launch are now defined in terms of each other, a re-launch test for user applications has been added and additional assertions around failing of 'launch' on a currently-launched Application.


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
Reviewed by @mmmulani IRL.

PR #205, opened 2016-02-09
Simplify Interaction API

lawrencelomax:
The Interaction API is getting a little complicated since there are a whole bag of methods that exist on categories. The Mutable Array of Interactions has also been removed and replaced with sequencing returning new interaction values.

Instead it makes more sense to produce subclasses of `FBSimulatorInteraction` that become specialized to a specific 'Application' or 'Agent'. This also makes it possible to define a `bind` method to compose interactions from other interactions. Meaning that in a future PR the following will be possible:

Instead of:
`[[[simulator.interact install:application] launchApplication:appLaunch] performInteractionWithError:&error]`
`[[[[simulator withLaunchConfiguration:appLaunch] install] launch] performInteractionWithError:&error]`

> 

The `withLaunchConfiguration` factory method can then pivot on the kind of the launch configuration and then return a 'LaunchInteraction' object that is specialized to the kind of launch. It also means that providing launches by a parameter and by a historical configuration have a common form:

`[[simulator lastLaunch] relaunch]`

`lastLaunch` calls `withLaunchConfiguration` with the last launch if one exists, or else returns an interaction that will always fail.


lawrencelomax:
@facebook-github-bot shipit

PR #206, opened 2016-02-10
Terminate Application

lawrencelomax:
Adds an API to terminate an application by a specific bundle ID. I want to revisit this later to ensure that that we specify 'interacting' with a process via bundle ids, app paths and pids as this will make the API far more flexible.


lawrencelomax:
@facebook-github-bot shipit

PR #207, opened 2016-02-11
Improve Listen Lifecycle

lawrencelomax:
The Listen Commands need a little polish:
- [x] Use the Event Reporting mechanism so events can be reported via JSON
- [x] Improve the Human Readable Output of some events
- [x] Log Signal Handling with Info about the Signal
- [x] Blackhole to ignored function with `signal`
- [x] Improve error handling for 'listens'
- [x] Don't log that a listen has started until ports are bound.
- [x] User Initiated help to distinguish status codes


lawrencelomax:
@facebook-github-bot shipit

PR #210, opened 2016-02-15
Don't fail to launch if Application was launched in the background

lawrencelomax:
This could be optional in future, or we should have an API for disabling background modes for applications.


lawrencelomax:
@facebook-github-bot shipit

PR #212, opened 2016-02-16
Add 'all' as query for targeting all Simulators

lawrencelomax:
There's currently no way of selecting 'all' simulators when performing a query, this means that killing or deleting all the simulators has to be done via pool preconditions like `--delete-all` and `--kill-all`. This doesn't make a huge amount of sense, as having purpose built operations is much more obvious. 

Also adds defaulting of `Query` for the `list` command so that `fbsimctl list` lists all Simulators.


lawrencelomax:
@facebook-github-bot shipit

PR #214, opened 2016-02-17
Chainable Actions

lawrencelomax:
Removing the distinction between `Interaction` and `Action` is worthwhile as it will allow actions to compose far easier. With the presence of `Defaults` to pass through 'the last Simulator we called' chains with booting and listening become easy:

`fbsimct create 'iPhone 5' boot --direct-launch --record-video listen --http 8090 shutdown diagnose`

This will create then boot a simulator, recording the Framebuffer's video, then listen on port 8090. When a `SIGTERM` is sent to the process, it will shutdown and print out the logs.

This should also make it a bit easier to write help.


lawrencelomax:
@facebook-github-bot shipit

PR #215, opened 2016-02-18
Fix argv[0] on agent launches to be the executable path

lawrencelomax:
This can confuse launched processes as the `argv[0]` isn't automatically set, it just uses the arguments array verbatim.


asm89:
watman

PR #216, opened 2016-02-20
Create FBSimulatorSet

lawrencelomax:
`FBSimulatorPool` was previously responsible for two things:
1) Managing the Creation/Erasing/Deletion of Simulators via `FBSimulatorConfiguration` with robustness.
2) Allocation: Allowing reservation of Simulators and options for teardown in a convenient API.

Now `FBSimulatorSet` does 1) and `FBSimulatorPool` does 2). This is useful as `fbsimctl` only really cares about a robust CRUD API and doesn't need the Allocation semantics whatsoever. This also frees up API users of `FBSimulatorControl` to ignore the Pool entirely and just go straight for the device set.


lawrencelomax:
@facebook-github-bot import

PR #217, opened 2016-02-20
Parsable as a Property

lawrencelomax:
`Parsable` is a protocol for providing a parser of a specific type. This was previously provided by a static method, but since swift supports static vars it makes more sense to use those.


lawrencelomax:
@facebook-github-bot shipit

PR #218, opened 2016-02-20
Rename performInteractionWithError: -> perform:

lawrencelomax:
Seems a bit silly to encode the reciever's protocol in the method name. There's no ambiguity with existing selectors in Foundation.


lawrencelomax:
@facebook-github-bot shipit

PR #219, opened 2016-02-21
Add FBProcessInteraction

lawrencelomax:
There are a whole bunch of ways of specifying a way of killing a process based on the different ways that a running process can be specified. By bringing these together in `FBProcessInteraction` the method that does the killing can be consolidated.

### DIFF
--- a/FBSimulatorControl/Interactions/FBSimulatorInteraction.m
+++ b/FBSimulatorControl/Interactions/FBSimulatorInteraction.m
@@ -90,23 +90,6 @@
   return [self interactWithSimulatorAtState:FBSimulatorStateBooted block:block];
 }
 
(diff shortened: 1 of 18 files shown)


FBSimulatorControl/Interactions/FBSimulatorInteraction.m
```diff
@@ -60,59 +60,42 @@ - (instancetype)copyWithZone:(NSZone *)zone
 
 #pragma mark Private
 
-- (instancetype)interactWithSimulator:(BOOL (^)(NSError **error, FBSimulator *simulator))block
+- (instancetype)interactWithSimulator:(BOOL (^)(id interaction, NSError **error, FBSimulator *simulator))block
```
marekcirkos: I think it would be better to actually define all interactions as `id<FBInteraction>` instead of `id`. I know it is not needed most of the time, but you lose compiler type checking for no reason.

lawrencelomax:
use the robots instead.

PR #220, opened 2016-02-21
Workaround 'Duplicate Protocol Conformance' Errors

lawrencelomax:
Looks like Swift doesn't like certain protocol extensions and will sporadically fail. This should fix that.


lawrencelomax:
@facebook-github-bot shipit

PR #222, opened 2016-02-22
Build ActionPerformer from CommandPerformer

lawrencelomax:
Previous iterations of any `Listen` mode have fixed the `Query` to the `Query` of the `Listen` command. This is less desirable than allowing this to be optionally changed.


lawrencelomax:
@facebook-github-bot shipit

PR #224, opened 2016-02-24
Fix build.sh copying

lawrencelomax:
The quoting means that the globbing of the artifact path fails.


lawrencelomax:
@facebook-github-bot shipit

PR #225, opened 2016-02-24
Don't fail on an empty pool

lawrencelomax:
Calling `fbsimct all delete` on an empty pool really shouldn't error. In fact all commands that operate on an empty pool should return fine with a logger message.

Additionally, failure messages weren't sent to the reporter, which would mean that some failure messages aren't json-ified.


lawrencelomax:
@facebook-github-bot shipit

PR #226, opened 2016-02-25
Additional Robustness in FBSimulatorSet

lawrencelomax:
- Add a backreference from `FBSimulatorSet` to `FBSimulatorControl`
- Don't assert on wrong sim, return error.
- Create `FBSimulatorSet` and pass-from-above to `FBSimulatorPool`


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
reviewed on Phab

PR #228, opened 2016-02-27
Scale at the SimDeviceFramebufferService level

lawrencelomax:
The current way of creating the `SimDeviceFramebufferService` uses the `mainScreen` resolution. When recording a video, this is then thrown away and downscaled, which is a waste of resources. A 50% scale is a 4X reduction in the Framebuffer size and means less bytes are going to be copied around, leading to a lower resource utilisation on the host.


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
Accepted on the phab.

PR #229, opened 2016-03-01
Monotonic Framebuffer Frame Times

lawrencelomax:
Instead of constructing the current time of a frame at the `FBFramebufferVideo`'s `mediaQueue`, we should construct it at the time that the `FBSimulatorFramebuffer` gets the frames on it's own queue, with increased precision.

This should give a more accurate time than on the `mediaQueue`, both in terms of resolution of the time (by using a timescale that matches the system clock) and because `CMTime` values are created from the timebase when frames are pushed to the `FBSimulatorFramebuffer`'s private queue. Time conversions to a timescale more suitable for video encoding can then occur in `FBFramebufferVideo`.

Since there are a boatload of values passed around on the `FBSimulatorFramebufferDelegate` methods the params are now put inside a `FBFramebufferFrame` object. This also aids with conversions between timebases and rounding modes.


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
Accepted on Phab

PR #230, opened 2016-03-01
Use MPEG4 Container

lawrencelomax:
Depending on #229, running the tests to confirm that the reliability of video encoding has improved now there is timebase conversion in #229.


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
Accepted on Phab

PR #231, opened 2016-03-01
Add Automatic and Manual Video Recording

lawrencelomax:
Prior to this PR, it starting and stopping of video recording was completely automatic and would be enabled/disabled per Simulator boot. This PR adds an start/stop recording API and an optional launch flag for enabling the automatic recording of video on Simulator boot.


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
Reviewed on phab.

PR #232, opened 2016-03-02
Provisional IPC Support

lawrencelomax:
Since a process that links `FBSimulatorControl` may contain the 'Framebuffer' for any given booted Simulator, a secondary process may wish to manipulate this Simulator. In this 'direct-launch' model a process needs to be kept alive to act as a semi-persistent that should stay alive as long as the Simulator must stay alive.

Terminating the process that owns a `SimDevice`'s `PurpleFBServer` will shutdown the Simulator. This can be seen in the `Simulator.app`. Since a process that contains the Simulator's Framebuffer can take the form of a `fbsimctl` process, it's possible to use one process to hold onto the booted Simulator's Framebuffer, and another to start the video recording:

- `fbsimctl UDID boot --direct-launch --wait` Will boot a Simulator and keep it alive in the `fbsimctl` process. Sending this process a `SIGTERM` will shutdown the Simulator gracefully. The Framebuffer will be connected to this Simulator, but recording will not start until recording is explicitly requested.
- `fbsimctl UDID record start` Will communicate to the above waiting process and start the video record. It will then terminate.

For example, this is useful when you wish to record the video of a Simulator, but want to start recording at a later date.


lawrencelomax:
Use the robots instead

PR #233, opened 2016-03-02
Add FBFramebufferVideoConfiguration

lawrencelomax:
Sadly, it looks like there are still issues with encoding when using `AVFileTypeMPEG4`. Instead of messing around with this in master until we find the sweet spot, it makes more sense to be able to manipulate the configuration of video encoding in clients that care about this. `FBFramebufferVideoConfiguration` takes on this role.

Additionally the `--record-video` flag disappears and can be replaced with `record start` by chaining interactions in `fbsimctl`.


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
Reviewed in phab

PR #234, opened 2016-03-03
FBConfigurationTests

lawrencelomax:
Adds a new `FBConfigurationTests` class that adds some tests for common 'value-object' scenarios as well as JSON Describeability.


lawrencelomax:
@facebook-github-bot shipit

PR #235, opened 2016-03-03
Rename FBSimulatorFramebuffer to FBFramebuffer

lawrencelomax:
All the the other Frambuffer classes (including the delegate) begin with `FBFramebuffer` so the container object should too.


lawrencelomax:
@facebook-github-bot shipit

PR #236, opened 2016-03-03
Record Start/Stop of video with JSON Payload

lawrencelomax:
Having separate endpoints for each action seems a little sad, so we can now parse numbers as bools and use this to flip the `Action.Record` action.

`NSJSONSerialization` represents Bools as `NSNumber` so all numbers can be interpreted as bools :/


lawrencelomax:
@facebook-github-bot shipit

PR #237, opened 2016-03-03
Add FBFramebufferVideoOptions

lawrencelomax:
There are a number of optional behaviours in `FBFramebufferVideo` that can be switched on and off independently, so they have been given an option Set.

- `FBFramebufferVideoOptionsAutorecord` takes the place of the `autorecord` BOOl on `FBFramebufferConfiguration`
- `FBFramebufferVideoOptionsImmediateFrameStart` will cause the prior frame event to be kept around so that the record can start 'immediately'. The reason for doing this is that if the Simulator is at the home screen the first frame that will be received after the 'waiting for frame' state is set will be changes to the clock (i.e. once per minute) since Springboard doesn't need to redraw until the clock changes.
- `FBFramebufferVideoOptionsFinalFrame` is like the above, but upon stopping the video. By re-pushing the last frame the video will show how long the Simulator sat at a particular 'still frame' for.

### DIFF
--- a/FBSimulatorControl/Configuration/FBFramebufferVideoConfiguration.h
+++ b/FBSimulatorControl/Configuration/FBFramebufferVideoConfiguration.h
@@ -16,7 +16,17 @@
 @class FBDiagnostic;
 
 /**
+ Options for FBFramebufferVideo.
+ */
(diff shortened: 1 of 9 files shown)


lawrencelomax:
@facebook-github-bot shipit

PR #239, opened 2016-03-06
Linking Info.plist to main framework target

davidlawson:
For some reason this was unlinked, causing some code signing issues.


lawrencelomax:
@facebook-github-bot shipit

PR #240, opened 2016-03-06
Removing shared schemes for fbsimctl so Carthage will only find FBSimulatorControl

davidlawson:
A release should also be tagged so Carthage can be used without specifying a branch.


bootstraponline:
If the schemes are removed then `fbsimctl` can't be built with carthage...

lawrencelomax:
Sorry for the delay... I wasn't aware that carthage would build from the `fbsimctl` scheme. If you just want the `FBSimulatorControl` framework, you should just be able to use the `FBSimulatorControl.xcodeproj`. Is this not sufficient for your needs?

davidlawson:
I must have removed it as it wasn't building.

`carthage update` with `github "facebook/FBSimulatorControl" "master"` doesn't build due to:
```
error: no such module 'Swifter'
```

lawrencelomax:
> carthage update with github "facebook/FBSimulatorControl" "master" doesn't build due to:

I guess I need to add a Cartfile to the repo to fix this then. I would only think that anybody would ever want to build and pull in `FBSimulatorControl` rather than the CLI target itself.

lawrencelomax:
This is way old and apoligies for closing it now. If this is an issue, can you re-open and we can track there.

PR #243, opened 2016-03-07
Common start/end events for all actions

lawrencelomax:
All Actions will now report start/end events to the `EventReporter`, this previously was defined per-action. Additionally each diagnostic is a separate event rather than a single event with an array subject, removing one level of nesting.


lawrencelomax:
@facebook-github-bot import

lawrencelomax:
@facebook-github-bot shipit

PR #244, opened 2016-03-07
FBLogSearch & FBBatchLogSearch

lawrencelomax:
Adds a `FBLogSearch` which is a simple Log Searching class that can search through the text of text files. This is particularly practical if you wish to extract information from an app log in order to assert against particular conditions.

The implementation is pretty naiive, but relies on `FBConcurrentCollectionOperations` to be 'fast-enough'. By using a class cluster, it should be possible to do log searching directly via asl in the future.

In addition to the 1-diagnostic 1-predicate search of `FBLogSearch` there is a complementary class `FBBatchLogSearch` which builds upon `FBLogSearch` and will run multiple searches concurrently and bucket them by diagnostic name. This can be very handy when you wish to search for a bunch of strings in an app/system log, or there's an interesting string that you want to track down across a range of files.

Additionally, the searches are de/serializable over JSON, which makes it trivial to implement and endpoint to take care of log searching. In some occasions it can be useful to query a remote simulator for particular information in the logs. Batching this all together in one call can make the query more performant.


lawrencelomax:
Going to use the robots instead.

PR #245, opened 2016-03-08
FBDiagnostic with bi-directional JSON Coercions

lawrencelomax:
JSON is used throughout `FBSimulatorControl` and is used widely for wiring data around in a structured way. For example the `tree` command on `WebDriverAgent` produces a JSON Tree of the element hierarchy. This is very helpful for diagnosing common issues when running end-to-end tests with `WebDriverAgent`. By being able to treat this as a 'Diagnostic' it becomes easier to shuffle it around.


lawrencelomax:
@facebook-github-bot shipit

PR #246, opened 2016-03-08
Fix FBDiagnostic File Writing

lawrencelomax:
There was a case for `FBDiagnostic` writing that meant that new data in old paths could result in the old data staying, so this has been fixed along with some unit tests for these cases.


lawrencelomax:
@facebook-github-bot shipit

PR #247, opened 2016-03-09
Add FBJSONDeserializable

lawrencelomax:
Since a bunch of value objects are capable of being defined in terms of a JSON-serialized version, some of them should be able to be de-serialized from the very same data structure. This simplifies and consolidates the responsibility for `fbsimctl`s HTTP Server. 

As endpoints are added to the HTTP Server, the required values to perform any action can then be constructed from wired JSON in the POST of the request's body. This also means that the knowledge of what these fields look like stays in `FBSimulatorControl`.

Additional test suites have been added to confirm that the shape of the json and the equality of the POOCOs holds.


lawrencelomax:
@facebook-github-bot shipit

PR #251, opened 2016-04-04
Fix iPad Pro configuration creation

SemenovAlexander:
This fixes issue, when iPad Pro configuration couldn't be created by name, because of missing in configurations array


lawrencelomax:
@facebook-github-bot shipit

PR #257, opened 2016-04-12
Fix fbsimctl build breaker

lawrencelomax:
The class owning this has changed.


lawrencelomax:
@facebook-github-bot shipit

PR #264, opened 2016-05-19
Fix setting up test connection on iOS 8 simulators

### LINKED ISSUE #263: Connection to test manager daemon times out on iOS 8.4 simulator
Hi,

thanks for the great work on this project, I'm following the changes every day with excitement :).

I'm playing around with `XCTest` execution on different simulators. It works great if I execute the test suite on `FBSimulatorConfiguration.iPhone5.iOS_9_3` or on `FBSimulatorConfiguration.iPhone6.iOS_9_3` for example. However the connection to the test manager daemon always times out when using `FBSimulatorConfiguration.iPhone5.iOS_8_4`. The error message looks like this:

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason:
'Error Domain=com.facebook.FBSimulatorControl Code=0 "Failed start test manager"
UserInfo={NSLocalizedDescription=Failed start test manager, NSUnderlyingError=0x114aa6820
{Error Domain=com.facebook.XCTestBootstrap Code=0 "Failed to connect to the daemon" UserInfo=
{NSLocalizedDescription=Failed to connect to the daemon, NSUnderlyingError=0x114aa93d0
{Error Domain=com.facebook.XCTestBootstrap Code=0 "Timed out waiting for daemon connection" UserInfo=
{NSLocalizedDescription=Timed out waiting for daemon connection}}}}}'
```

If you need more information, please let me know!

I'm on this commit btw: 2159e739525152c67882cadd94dcb95976470335

/Edit: And on Xcode 7.3.1

Thanks,
Johannes

plu:
Fix #263

Please let me know if I can improve something. Please see also my comments in #263 about the retrying strategy of Xcode. Once you "listen" to the `XPC` communication, you'll notice that Xcode is also just trying different methods and ignoring errors.

### DIFF
--- a/XCTestBootstrap/TestManager/FBTestDaemonConnection.m
+++ b/XCTestBootstrap/TestManager/FBTestDaemonConnection.m
@@ -186,10 +186,10 @@
   [receipt handleCompletion:^(NSNumber *version, NSError *error) {
     if (error) {
       [self.logger logFormat:@"Error in whitelisting response from testmanagerd: %@ (%@), ignoring for now.", error.localizedDescription, error.localizedRecoverySuggestion];
-      return;
+    } else {
+      self.daemonProtocolVersion = version.integerValue;
+      [self.logger logFormat:@"Got legacy whitelisting response, daemon protocol version is 14"];
     }
-    self.daemonProtocolVersion = version.integerValue;
-    [self.logger logFormat:@"Got legacy whitelisting response, daemon protocol version is 14"];
     self.state = FBTestDaemonConnectionStateReadyToExecuteTestPlan;
   }];
   return receipt;
--- a/FBSimulatorControlTests/Tests/Integration/FBSimulatorTestInjectionTests.m
+++ b/FBSimulatorControlTests/Tests/Integration/FBSimulatorTestInjectionTests.m
@@ -51,6 +51,18 @@
 
 }
 
+- (void)testInjectsApplicationTestIntoSampleAppOnIOS81Simulator
+{
+  self.simulatorConfiguration = FBSimulatorConfiguration.iPhone5.iOS_8_1;
+  FBSimulator *simulator = [self obtainBootedSimulator];
+  id<FBInteraction> interaction = [[simulator.interact
+    installApplication:self.tableSearchApplication]
+    startTestRunnerLaunchConfiguration:self.tableSearchAppLaunch testBundlePath:self.applicationTestBundlePath reporter:self];
+
+  [self assertInteractionSuccessful:interaction];
+  [self assertPassed:@[@"testIsRunningOnIOS"] failed:@[@"testIsRunningOnMacOSX", @"testIsSafari"]];
+}
+
 - (void)testInjectsApplicationTestIntoSafari
 {
   FBSimulator *simulator = [self obtainBootedSimulator];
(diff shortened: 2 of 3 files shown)


lawrencelomax:
@plu can you rebase on master and re-try. @marekcirkos recently pushed some commits that fixes the state change for the legacy protocol.

plu:
Did a rebase and added some test, so we notice when test execution would break on iOS 8. What @marekcirkos did only fixes part of the issue. You can try executing `testInjectsApplicationTestIntoSampleAppOnIOS81Simulator` from f877e5d without the changes I did in  d370c69 - it will fail. Only after applying d370c69 it passes. The older Simulator runtime expects `_IDE_beginSessionWithIdentifier` instead of `_IDE_initiateSessionWithIdentifier `. We also need to change the early returns in case of an error and just log the error instead.

XCTestBootstrap/TestManager/FBTestDaemonConnection.m
```diff
@@ -186,10 +186,10 @@ - (DTXRemoteInvocationReceipt *)setupDaemonConnectionViaLegacyProtocol
...
     if (error) {
       [self.logger logFormat:@"Error in whitelisting response from testmanagerd: %@ (%@), ignoring for now.", error.localizedDescription, error.localizedRecoverySuggestion];
-      return;
+    } else {
+      self.daemonProtocolVersion = version.integerValue;
+      [self.logger logFormat:@"Got legacy whitelisting response, daemon protocol version is 14"];
     }
-    self.daemonProtocolVersion = version.integerValue;
-    [self.logger logFormat:@"Got legacy whitelisting response, daemon protocol version is 14"];
     self.state = FBTestDaemonConnectionStateReadyToExecuteTestPlan;
```
marekcirkos: Probably we should introduce error state to handle errors properly. Otherwise we will mask errors.

marekcirkos:
Cool. We didn't intend to support versions pre iOS 8.0. It is cool ti works with so simple change.

marekcirkos:
@facebook-github-bot shipit

XCTestBootstrap/TestManager/FBTestDaemonConnection.m
plu: Yes, that makes sense. It's a bit strange that Xcode itself is also just trying the different `XPC` messages and is ignoring the errors coming back. We could be a bit smarter than that and check what Simulator version is being used and depending on that we select the correct `_IDE...` methods to invoke, instead of try-and-error :).

facebook-github-bot-5:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1774332506118072/int_phab) to review.

plu:
Thank you for the review & merge!

XCTestBootstrap/TestManager/FBTestDaemonConnection.m
marekcirkos: Trick is XCTestBootstrap needs to work for devices as well ;). Doing so would be easier when we finish unifying FBSimulatorControl and FBDeviceControl

plu: Ah - ok :). Must be exciting to work on this full time :).

Issue #281, opened 2016-07-26
fbsimctl: Support waitUntilAllTestRunnersHaveFinishedTestingWithTimeout


plu:
It would be great to support executing tests on the command line and wait for them to finish. Example:

```
fbsimctl --simulators $UDID launch_xctest $APP/PlugIns/UnitTests.xctest com.ebaykleinanzeigen.ebc --wait-with-timeout 600
```

Another thing that would be nice is getting the test output in some sort of machine readable format (JSON?). So we can transform that in whatever we like (e.g. JUnit).

lawrencelomax:
Yes! This is something that makes a lot of sense. It should be pretty easy to get this up and running. The APIs for waiting as you point out. It's just a case of parsing the arguments, writing a reporter, and passing this down into where the tests are launched.

This is a decent and fun task for anybody who wants to hack on `fbsimctl` 😄

plu:
I see that `fbsimctl` can already produce JSON formatted output. I have the feeling that implementing a generic reporter would make more sense. Then anybody can use that JSON stream as an input and transform it to whatever they like, e.g. JUnit. What do you think?

Regarding wiring `waitUntilAllTestRunnersHaveFinishedTestingWithTimeout`: I assume I should extend `Action.LaunchXCTest` in some way so that I can pass in a timeout value, right?

lawrencelomax:
> I see that fbsimctl can already produce JSON formatted output. I have the feeling that implementing a generic reporter would make more sense

Yes, there is a reporter that is used to report events in either a human readable or JSON format. I never really envisioned using it for reporting test results however, as the format of the reported output can be substantially different between `JUnit`, `RSpec`, whatever. I would have thought going directly to the target output would make it easier to integrate with other tools, rather than adding another translation layer.

I could see that you could use the current reporter as-is, by passing it to a concrete implementation of `FBTestManagerTestReporter` that takes the `fbsimctl` reporter could be a good way to get started. You could also put the reporter in `XCTestBootstrap` and then it can be used by other consumer libraries.

> Regarding wiring waitUntilAllTestRunnersHaveFinishedTestingWithTimeout: I assume I should extend Action.LaunchXCTest in some way so that I can pass in a timeout value, right?

Yep, you can just add another associated type to the enumeration. I'd consider making the timeout optional for the time being. There's plenty of examples for how to construct a flag like this in `CommandParsers` and `CommandParsersTests`.

plu:
[...]

PR #290, opened 2016-08-07
Add JUnit reporter

plu:
In #281 we discussed about additional test reporters. Here comes the JUnit one. In `xctool` I saw that they kind of flatten the nested test suites. This looks better on the CI report especially if you have mixed Objective-C and Swift test cases in your test target. But in the end I think supporting nested test suites is the way to go. Maybe I will add a `BOOL flattenTestSuites` later.


plu:
I have a bit mixed feelings about the naming of the classes, that I created in this pull request. I believe that following classes might be shared by other test reporters:

* FBTestManagerTestReporterTestCase
* FBTestManagerTestReporterTestCaseFailure
* FBTestManagerTestReporterTestSuite

Hence they are missing the `JUnit` part in their name.

lawrencelomax:
😃  Looks like a lot of thought and effort went into this.
Might take a while for me to review this one, but will try to get around to it today.

plu:
Thanks! I'm in the middle of the `RSpec` formatter. Maybe it'd make sense for you waiting for that, I noticed it makes sense to extract some common things there.

XCTestBootstrapTests/FBTestManagerTestReporterJUnitTests.m
```diff
@@ -0,0 +1,111 @@
...
+{
+  [super setUp];
+
+  self.testManagerAPIMediator = [OCMockObject mockForClass:[FBTestManagerAPIMediator class]];
+  self.outputFilePath = [NSTemporaryDirectory() stringByAppendingPathComponent:[NSUUID UUID].UUIDString];
+
+  [[NSFileManager defaultManager] createFileAtPath:self.outputFilePath contents:nil attributes:nil];
+
+  self.outputFileHandle = [NSFileHandle fileHandleForWritingAtPath:self.outputFilePath];
+  self.reporter = [FBTestManagerTestReporterJUnit withOutputFileHandle:self.outputFileHandle];
```
lawrencelomax: It looks like you never close `outputFileHandle` which means you'll leak file descriptors in these tests.

XCTestBootstrap/TestManager/FBTestManagerTestReporterJUnit.h
```diff
@@ -0,0 +1,30 @@
...
+ It writes the Test Result to a given File Handle in the JUnit XML format.
+ */
+@interface FBTestManagerTestReporterJUnit : NSObject <FBTestManagerTestReporter>
+
+/**
+ Constructs a JUnit Test Reporter.
+
+ @param outputFileHandle a file handle the JUnit XML should be written to.
+ */
++ (instancetype)withOutputFileHandle:(NSFileHandle *)outputFileHandle;
```
lawrencelomax: Is the main reason for using a file handle here that you can stream the results to a pipe instead of a file?

It does add a burden to the consumer of the API to ensure that the file handle is closed. If you only care about files you could just give a file path and manage the file handle internally to the class. The reporter could also clean the file handle on `dealloc` or when there is a final event in the `FBTestManagerTestReporter` protocol indicating that the stream should be closed.

XCTestBootstrapTests/FBTestManagerTestReporterJUnitTests.m
lawrencelomax: Ignore that, I just looked at the docs and `NSFileHandle` will close the resource when it is deallocated.

XCTestBootstrap/TestManager/FBTestManagerTestReporterJUnit.h
lawrencelomax: Add an `@return` in the documentation comment.

XCTestBootstrap/TestManager/FBTestManagerTestReporterJUnit.m
```diff
@@ -0,0 +1,185 @@
...
+        finishedWithSummary:(FBTestManagerResultSummary *)summary
+{
+  self.currentTestSuite.summary = summary;
+  self.currentTestSuite = self.currentTestSuite.parent;
+}
+
+- (void)testManagerMediatorDidFinishExecutingTestPlan:(FBTestManagerAPIMediator *)mediator
+{
+  NSXMLDocument *document = [FBTestManagerTestReporterJUnit documentForTestSuites:self.rootTestSuites];
+  [self.outputFileHandle writeData:[document XMLDataWithOptions:NSXMLNodePrettyPrint]];
```
lawrencelomax: It looks like you're only using the file handle to write the data in one big batch at the end. If you only want to support writing to file, you can simplify the API by just passing a file path in the constructor and avoid the need to pass a file handle.

XCTestBootstrap/TestManager/FBTestManagerTestReporterTestCase.h
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ The test report status.
+ */
+@property (nonatomic) FBTestReportStatus status;
+
+/**
+ The test case duration.
+ */
+@property (nonatomic) NSTimeInterval duration;
```
lawrencelomax: Make this externally `readonly` if it's not for consumers to modify. You can add a method like `finishWithDuration:status:` that can instead be called from `FBTestManagerTestReporterJUnit`. This can place additional requirements that the  `finishWithDuration:status:` is called only once per-instance.

XCTestBootstrap/TestManager/FBTestManagerTestReporterTestCase.m
```diff
@@ -0,0 +1,72 @@
...
+
+#pragma mark -
+
+- (NSString *)description
+{
+  return [NSString stringWithFormat:@"TestCase %@ | Method %@ | Status %@ | Duration %f", self.testClass, self.method,
+                                    self.statusDescription, self.duration];
+}
+
+- (NSString *)statusDescription
```
lawrencelomax: Add the inverse of the `+[FBTestManagerResultSummary statusForStatusString:]` method and call it from here.

XCTestBootstrap/TestManager/FBTestManagerTestReporterTestCase.h
```diff
@@ -0,0 +1,64 @@
...
+ @param testClass the test class name.
+ @param method the test method name.
+ @return a new test case instance.
+ */
++ (instancetype)withTestClass:(NSString *)testClass method:(NSString *)method;
+
+/**
+ The test report status.
+ */
+@property (nonatomic) FBTestReportStatus status;
```
lawrencelomax: add `assign`. Also see the `readonly` comment below.

XCTestBootstrapTests/FBTestManagerTestReporterJUnitTests.m
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+#import <OCMock/OCMock.h>
+#import <XCTest/XCTest.h>
+
+#import "FBTestManagerAPIMediator.h"
+#import "FBTestManagerResultSummary.h"
+#import "FBTestManagerTestReporterJUnit.h"
+
+#define testSuiteDidStart(name, startTime)                                                                             \
```
lawrencelomax: Is there a good reason for using Macros here? You could define each of these as instance methods of the test case and get the same effect. Macros will make this harder to debug when a test fails since all of the parameters for the assertions will be inlined.

lawrencelomax:
Having a base `Reporter` model for accumilating test state makes a lot of sense here. There are a few comments that need addressing but the approach looks sound.

You may also want to consider writing an integration or end-to-end test for this. [`FBSimulatorTestInjectionTests`](https://github.com/facebook/FBSimulatorControl/blob/master/FBSimulatorControlTests/Tests/Integration/FBSimulatorTestInjectionTests.m) shows what this make look like. `fbsimctl` also has a suite of full [end-to-end tests in python](https://github.com/facebook/FBSimulatorControl/blob/master/fbsimctl/cli-tests/tests.py) that will allow you a place to assert based on the output. We have a `SimpleTestTarget.xctest` target that's a fixture of the `FBSimulatorControl` project that will provide you a full test target for running against.

plu:
Thank you for the detailed review! I will follow up and address all the issues you mentioned.

XCTestBootstrap/TestManager/FBTestManagerTestReporterJUnit.h
plu: For the `RSpec` reporter you probably want to write to `stdout` or `stderr`, I assume. I gave the `JUnit` reporter the same interface so later it will be easier to plug that into `fbsimctl`. I had an interface similar to this in mind:

```
fbsimctl --reporter=rspec # writes to stdout
fbsimctl --reporter=junit junit.xml # write to the file junit.xml
```

If all reporters share the same interface (init with a file handle), it will be easier. But on the other hand you will probably never want to write the `JUnit` output to `stdout` or `stderr`, but always to a file. I guess I should change the interface then.

plu:
I have addressed all your comments, thanks again for taking the time for the review :).

When I was running the test suites of different projects I have noticed that there will always be only one root test suite. So I have adopted the implementation to that. The bits of collecting the test result data I have extracted in a common base class called `FBTestManagerTestReporterBase`, which the `RSpec` reporter can inherit from then too.

XCTestBootstrapTests/FBTestManagerTestReporterJUnitTests.m
```diff
@@ -0,0 +1,169 @@
...
+  NSURL *fixtureFileURL = [NSURL fileURLWithPath:[FBTestManagerTestReporterJUnitTests JUnitXMLResult0Path]];
+  NSString *actual = [self stringWithContentsOfJUnitResult:self.outputFileURL];
+  NSString *expected = [self stringWithContentsOfJUnitResult:fixtureFileURL];
+
+  XCTAssertEqualObjects(expected, actual);
+}
+
+- (void)testJUnitReporterWithManyNestedTestSuites
+{
+  [self testSuite:@"One" didStartAt:@"2016-08-07 10:31:33"];
```
lawrencelomax: 👍👍

XCTestBootstrap/TestManager/FBTestManagerTestReporterTestCase.m
```diff
@@ -0,0 +1,72 @@
...
+}
+
+- (void)addFailure:(FBTestManagerTestReporterTestCaseFailure *)failure
+{
+  [self.mutableFailures addObject:failure];
+}
+
+- (void)finishWithStatus:(FBTestReportStatus)status duration:(NSTimeInterval)duration
+{
+  NSAssert(!self.finished, @"finishWithStatus:duration: may be called only once");
```
lawrencelomax: 👍

XCTestBootstrap/TestManager/FBTestManagerResultSummary.h
```diff
@@ -31,8 +31,22 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign, readonly) NSTimeInterval testDuration;
 @property (nonatomic, assign, readonly) NSTimeInterval totalDuration;
 
+/**
+ Returns a status enum value for the given status string.
+
+ @param statusString the status string.
```
lawrencelomax: Nice catch!

lawrencelomax:
This is really nice work. I know you've seen the `fbxctest` diff that's landed. Obviously we will eventually want to migrate all of this reporting to `fbxctest`, but the work you've done here can be re-used in `fbxctest`.

@facebook-github-bot import

facebook-github-bot-6:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3689915) to review internal test results.

plu:
Makes totally sense to have a different CLI for controlling the Simulators and for executing the tests.

Regarding reporters: Next I'll look into the `RSpec` formatter.

lawrencelomax:
I've ran the our linter over your PR and there were a few changes. You'll have to rebase when it lands, but there's nothing that should be too surprising.

plu:
Don't worry about it! Thanks for reviewing/importing it :)!

PR #292, opened 2016-08-08
Add output/ and .clang-format to gitignore

plu:
* `output/` is created by `build.sh`
* I'm using some `.clang-format` configuration to achieve a similar code formatting like the one the project has already in existing code. Do you guys maybe have such a configuration too that you could share, so it wouldn't just be "similar" but exactly the same?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,5 @@ Carthage/Checkouts
 fbsimctl/Carthage/Build
 fbsimctl/Carthage/Checkouts
 
+output/
+.clang-format


lawrencelomax:
(y)

We should probably rename `output` as well, or use a dot-prefixed directly as it is a bit vague. 

I don't have a clang format configuration I use, we do have a linter that will trigger when I import commits to Phabricator, but it only has a few formatting rules. If you have one to share I'd certainly take a look at it.

@facebook-github-bot shipit

facebook-github-bot-0:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3683549) to review internal test results.

PR #299, opened 2016-08-14
Extract FBXCTestKit.framework

plu:
* Extract FBXCTestKit.framework
* Add simple integration test
* Support building and testing `fbxctest` via build script
* Build and test `fbxctest` on Travis


plu:
I didn't change much of the code, yet. Mainly just adding a test target, moving things around and extending the build script.

lawrencelomax:
Thanks so very much for this!

I was looking at doing something very similar so that we had a target to run integration tests against, as well as a Framework bundle container for any resources or shims that need to be bundled in order to make the cli work. I also want to bundle more test target fixtures in order that we can validate behaviour against different kinds and platforms of test targets. 

Since this changes the build script I'll need to update the homebrew formula to run the new fbsimctl build command.

Let me know when it's ready for review.

plu:
I think it's fine to be reviewed now.

Unfortunately I'm a bit confused about `FBXCTestReporter` and how to get `FBTestManagerTestReporterComposite` and `FBTestManagerTestReporterJUnit` into `fbxctest` now. That's what I'd have liked to work on next, supporting multiple `-reporter` arguments.

lawrencelomax:
@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3716034) to review internal test results.

build.sh
```diff
@@ -102,11 +102,11 @@ function all_frameworks_test() {
 function strip_framework() {
   local FRAMEWORK_PATH="$BUILD_DIRECTORY/Build/Products/Debug/$1"
   echo "Stripping Framework $FRAMEWORK_PATH"
-  rm -r "$FRAMEWORK_PATH"
+  rm -rf "$FRAMEWORK_PATH"
```
lawrencelomax: I doubt that you would want to use `-f`! Could lead to some really nasty behaviour.

fbxctest/FBXCTestKit/FBXCTestBootstrapper.h
```diff
@@ -0,0 +1,9 @@
+// Copyright 2004-present Facebook. All Rights Reserved.
+
+#import <Foundation/Foundation.h>
+
+@interface FBXCTestBootstrapper : NSObject
```
lawrencelomax: We should change the name of this class as it creates ambiguity with the `XCTestBootstrap` Framework. This may become less significant in the future if we decide to merge `XCTestBootstrap` & `FBXCTestKit` into `XCTestControl`

build.sh
plu: I've added it so it does not fail (due to `set -e`) if you try to remove something that does not exist. This way `cli_build` can be shared between `fbsimctl` and `fbxctest`. Otherwise we have to provide a separate one for each of them, because in `fbxctest` you won't have to strip anything in `FBSimulatorControlKit.framework` because that's just not being built for that target.
If we still want to share `cli_build`, but dislike `-f`, I can make the `strip_framework` check for existance of given path, before removing it.
What do you think?

lawrencelomax:
@facebook-github-bot import

facebook-github-bot-8:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3716034) to review internal test results.

PR #303, opened 2016-08-17
Fix launchctl path

lawrencelomax:
The path of the `launchctl` was erroneously using a hardcoded path, relative to the root of the xcode install, rather than the `SimRuntime`'s root. This changes that, so that we can always fetch the appropriate path for the executable.


lawrencelomax:
@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3728569) to review internal test results.

PR #304, opened 2016-08-17
Add Xcode 8 to Travis

lawrencelomax:
Starting by trying to build everything. I'll have to probably exclude certain variants from the build matrix so that we only run the builds that we know to be green. Swift Version and `xctool` may only permit certain builds to run on Xcode 8.


lawrencelomax:
Closing in favour of #313

PR #308, opened 2016-08-23
Feature/fix start test

sapieneptus:
WIP

Created alternative 'strategy' for start test which parallels `ideviceteststarter` approach: Assume you already have a DeviceAgent installed and assume you already know the session_id. Given that, simply make the umpteen-way handshake with testmanagerd. 

@jmoody - I think this will work, but I am unable to test it. It can't find the frameworks? I am not sure what's wrong, the framework search paths seemed ok, I'm not sure what changed. Obviously it passes in CI - does the CI machine have the FB frameworks installed globally or something? Would love to spend a moment debugging this, I have a high degree of confidence that `start_test` will work after this and the companion PR in `iOSDeviceManager`


sapieneptus:
@lawrencelomax sorry about this, there seems to be a git bug when we create a PR on our own fork it also makes one on master :/ Closing, please ignore.

PR #323, opened 2016-10-15
Assert Xcode version when building fbsimctl

plu:
Provide better error message if somebody installs via Homebrew and does not have Xcode 8 installed.


lawrencelomax:
Awesome! Thanks for making this generic, will be handy for migrations in the future

PR #334, opened 2016-11-18
Fix fbsimctl tests

plu:
Due to the new explicit default of `FBSimulatorBootOptionsAwaitServices` some tests had to be adjusted. Also the environment variable injection in `launch_xctest` broke at some point.


lawrencelomax:
Cheers dude (y)

PR #338, opened 2016-11-21
Expose editPropertyListFileRelativeFromRootPath

plu:
Edit a property list file of the Simulator by manipulating the content of the dictionary in the block.

The `setupKeyboard` method is changing a few of the keyboard settings, but not all of them. With this new method consumers of the Framework have full control of editing property list files that belong to the Simulator. Without the need of FBSimulatorControl having to know what property list files exist and what options they support.


lawrencelomax:
Awesome, Thanks!

PR #343, opened 2016-12-01
Fix parsing finishTime NSString into NSDate

plu:


plu:
On a second thought, should we maybe remove the attempt of parsing into `NSDate` and use the `NSString` we got instead? Parsing the date might also break between Xcode versions, if they change the format in any way. What do you think?

plu:
Please don't merge, I'm trying to see on Travis where the next commit I'm pushing will break (Xcode 8?, 7?).

plu:
Travis came back with:

* Xcode 8
** `-[FBSimulatorTestInjection testInjectsApplicationTestIntoSafari] : ((summary.finishTime) != nil) failed`
** `-[FBSimulatorTestInjection testInjectsApplicationTestIntoSampleApp] : ((summary.finishTime) != nil) failed`

* Xcode 7
** `-[FBSimulatorTestInjection testInjectsApplicationTestIntoSafari] : ((summary.finishTime) != nil) failed`

So what do you think of removing the date time parsing and just returning the string?

lawrencelomax:
> Parsing the date might also break between Xcode versions, if they change the format in any way

You could have a factory method for an `NSDateFormatter` that would produce the appropriate date formatter for the current version of Xcode. All of this is known in the `FBControlCoreGlobalConfiguration`, so you could use the `NSDecimalNumber` in there.

> On a second thought, should we maybe remove the attempt of parsing into NSDate and use the NSString we got instead

Sure we could do this. I think an `NSDate` is a more convenient API, as the consumer of this API can transform it into any date format that they wish. Is there only a difference between Xcode7 and Xcode8? If this is the case a simple conditional to give a different date formatter should do the trick.

If this changes in Xcode9, then we can assess then!

plu:
Now after looking closer into it, I believe that the `dateFormat` was not correct. So it might not even be an Xcode 8/7 issue.

Pull request is ready for review, unless Travis comes back with bad news.

lawrencelomax:
Looks good, I don't think we'll need to cache the `NSDateFormatter` yet since the Summary is created once per test run. We can always improve this in the future however.

PR #346, opened 2016-12-05
Fix test expectations in FBSimulatorTestInjectionTests

plu:


lawrencelomax:
Nice. Missed this one when I made the new test method. Thanks for getting the tests in a better state!

PR #347, opened 2016-12-06
Expose method to modify defaults

plu:
Expose `modifyDefaultsInDomainOrPath:defaults:error:` to allow subclassing FBDefaultsModificationStrategy.


plu:
Just pulled the `defaults write` change into my local setup here. It's awesome, I'm so happy that changing settings is now working with a running Simulator 👍 !

lawrencelomax:
I did remove the block based one since it wasn't the best API for general consumption, this API makes a lot more sense to expose. Pleased to hear it's working out for you!

PR #350, opened 2016-12-11
Fix FBDeviceTestRunStrategyTest

plu:


plu:
Already fixed via #352

PR #353, opened 2016-12-12
Fix loading libShimulator.dylib from wrong bundle in testAppCrashLogIsFetched test case

toshi0383:
The test case was trying to load libShimulator.dylib from wrong bundle.
This PR resolves this issue.

Note that the test is still failing with error: `Failed to find haystack log`
Any idea for this?

### DIFF
--- a/FBSimulatorControlTests/Tests/Integration/FBSimulatorDiagnosticsTests.m
+++ b/FBSimulatorControlTests/Tests/Integration/FBSimulatorDiagnosticsTests.m
@@ -43,7 +43,9 @@
   }
 
   FBSimulator *simulator = [self assertObtainsBootedSimulator];
-  FBApplicationLaunchConfiguration *appLaunch = [self.tableSearchAppLaunch.injectingShimulator withEnvironmentAdditions:@{@"SHIMULATOR_CRASH_AFTER" : @"1"}];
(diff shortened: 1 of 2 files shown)


FBSimulatorControlTests/Tests/Integration/FBSimulatorDiagnosticsTests.m
```diff
@@ -43,7 +43,9 @@ - (void)testAppCrashLogIsFetched
   }
 
   FBSimulator *simulator = [self assertObtainsBootedSimulator];
-  FBApplicationLaunchConfiguration *appLaunch = [self.tableSearchAppLaunch.injectingShimulator withEnvironmentAdditions:@{@"SHIMULATOR_CRASH_AFTER" : @"1"}];
+  NSString *path = [[NSBundle bundleForClass: self.class] pathForResource:@"libShimulator" ofType:@"dylib"];
```
lawrencelomax: Is the `injectingShimulator` method working without needing to make this change? The Change to the `pbxproj` should be sufficient.

toshi0383: Unfortunately it's not.
This change is also needed.

toshi0383:
🚢

PR #356, opened 2016-12-14
Support booting persistent Simulators

plu:


lawrencelomax:
Is the rationale here that you want a close analog to how `simctl` works?

plu:
Yep, that's correct.

lawrencelomax:
Closing with @plu's comment on the other PR

PR #357, opened 2016-12-15
Restart SpringBoard in overrideWatchDogTimer

plu:
Unfortunately when the Simulator is booted already, overriding the watch
dog timer does not have an effect immediately. Restarting the
SpringBoard makes it reload the defaults and see the overridden
watch dog timer.

### DIFF
--- a/FBSimulatorControl/Utility/FBSimulatorLaunchCtl.m
+++ b/FBSimulatorControl/Utility/FBSimulatorLaunchCtl.m
@@ -141,38 +141,7 @@
   return serviceName;
 }
 
-- (BOOL)restartServiceWithName:(NSString *)serviceName timeout:(NSTimeInterval)timeout error:(NSError **)error
-{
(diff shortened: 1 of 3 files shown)


FBSimulatorControl/Utility/FBSimulatorLaunchCtl.m
```diff
@@ -141,8 +141,39 @@ - (nullable NSString *)startServiceWithName:(NSString *)serviceName error:(NSErr
   return serviceName;
 }
 
+- (BOOL)restartServiceWithName:(NSString *)serviceName timeout:(NSTimeInterval)timeout error:(NSError **)error
```
lawrencelomax: Could you not use ` - (BOOL)amendRelativeToPath:(NSString *)relativePath defaults:(NSDictionary<NSString *, id> *)defaults managingService:(NSString *)serviceName error:(NSError **)error` to do this instead?

plu:
Maybe it's time to get glasses! Thanks for pointing me in the right direction :).

PR #363, opened 2017-01-21
Add error code for startup timeout

plu:
This lets consumers react to a timeout that happened during test
startup.


lawrencelomax:
I think some of these error codes are derived from internal error codes used by Apple's Frameworks.  Regardless it doesn't look like their value is at-all significant to the logic of the test, so there's harm in adding another one.

PR #365, opened 2017-01-24
Fix usage of lib shimulator

plu:
Add libShimulator.dylib to the Bundle Resources of FBControlCore. Recompile the binary with Xcode 8.2 and code sign it to fix injection.


lawrencelomax:
Thanks!

PR #367, opened 2017-01-25
Add configuration for iOS 10.3

jonstoneman:


lawrencelomax:
I think we raced to get this out and it was added with the other TV/Watch runtimes in 34d5ecb. Thanks for putting this up!

PR #366, opened 2017-01-24
Fix createStd(Err|Out)DiagnosticForSimulator

plu:

PR #369, opened 2017-01-26
Make log file names unique across multiple application launches

plu:
When launching the same application multiple times, the log file gets overwritten multiple times, because the name is not unique.

### DIFF
--- a/FBSimulatorControl.xcodeproj/project.pbxproj
+++ b/FBSimulatorControl.xcodeproj/project.pbxproj
@@ -11,6 +11,7 @@
 		05046C021D48B8BA00295EE1 /* FBTestManagerTestReporterComposite.m in Sources */ = {isa = PBXBuildFile; fileRef = 05046C001D48B8BA00295EE1 /* FBTestManagerTestReporterComposite.m */; };
 		051E96FC1E5544A200301DFF /* libShimulator.dylib in Resources */ = {isa = PBXBuildFile; fileRef = AA017F4C1BD7784700F45E9D /* libShimulator.dylib */; };
 		05471D691D5991AF00232CB2 /* junitResult0.xml in Resources */ = {isa = PBXBuildFile; fileRef = 05471D681D5991AF00232CB2 /* junitResult0.xml */; };
+		058663DD1E59AE080073699C /* libShimulator.dylib in Resources */ = {isa = PBXBuildFile; fileRef = AA017F4C1BD7784700F45E9D /* libShimulator.dylib */; };
 		0593EA101E371E5800EBD7CB /* libShimulator.dylib in Resources */ = {isa = PBXBuildFile; fileRef = AA017F4C1BD7784700F45E9D /* libShimulator.dylib */; };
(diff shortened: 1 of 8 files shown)


plu:
I decided to create a sub directory inside the storage directory so the file name is easier on the eye, at the same time ensuring uniqueness by moving it in a unique subdirectory of the storage directory. Hope that's okay?

lawrencelomax:
This looks good, but you will need to modify the `-[FBSimulatorDiagnostics stdOutErrDiagnostics]` method and others. This should be pretty simple, you just need to find non-empty files within the subtree that you're constructing.

plu:
Thanks for the pointer! I'll have another look at this one after #366 has landed. They are interrelated and finishing this one is easier after #366 is done.

lawrencelomax:
@plu are you able to make progress on this now #366 is merged?

plu:
I have added a test (7312628), which still passes after the change in 2edfd0a:

```objc
  [self assertFindsNeedle:@"Shimulator" fromHaystackBlock:^ NSString * {
    NSString *haystack = @"";
    for (FBDiagnostic *diagnostic in simulator.simulatorDiagnostics.stdOutErrDiagnostics) {
        haystack = [haystack stringByAppendingString:[diagnostic asString]];
    }
    return haystack.length ? haystack : nil;
  }];
```

I cannot find the connection between `diagnosticForProcessInfo:logBuilder:` and `stdOutErrDiagnostics`. 

* `diagnosticForProcessInfo:logBuilder:` seems to be writing a file to disk (via `asl`) once you invoke the method, which is now unique after my change.
* `stdOutErrDiagnostics` is looking for files that were generated via `[FBSimulatorDiagnostics stdOut:]` and `stdErr:`. These are still not unique and get overwritten every time, it seems. Maybe we should also fix that? Then We'll have to touch `stdOutErrDiagnostics` etc.

plu:
By the way, one more thought about pid vs UUID: Now we create truely unique log files, to avoid clashes with process identifiers. But when invoking `asl_set_query` we still have to use the pid in the query. So this might fetch logs from n launches (which had the same pid), putting them into two different places (because of the UUID), with the same content.

plu:
I'm not sure which place actually should be responsible for making a unique `storageDirectory`, for now I have put it here: 562c7c9

Another place that I had on the list was here: `[FBProcessLaunchConfiguration createDiagnosticForSelector:simulator:diagnosticOut:error:]`

What do you think @lawrencelomax?

FBSimulatorControl.xcodeproj/project.pbxproj
```diff
@@ -3191,6 +3192,7 @@
 			isa = PBXResourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				051E96FC1E5544A200301DFF /* libShimulator.dylib in Resources */,
```
plu: I'm not sure why this was missing again...

plu:
It would be cool if we could run some of the tests in `FBSimulatorDiagnosticsTests` on Travis, right now most of them are excluded.

plu:
I don't know what's happening in the project, but `libShimulator.dylib` keeps disappearing from the `FBControlCore` bundle resources. I added it twice in this pull request.

plu:
This is ready, please double check if 1920107 looks better now.

Issue #368, opened 2017-01-25
CLI exits with timeout error when booting iOS 8.1 on macOS 10.12


AndreyBelym:
If you try to start iOS 8.1 device e.g. iPhone 4s on macOS 10.12 Siera. `fbsimctl` will successfully start the simulator, and then will hang for some time (about a couple of minutes) and will exit with the following error:
```
Error Domain=com.facebook.FBSimulatorControl Code=0 "Timed out waiting for all required services [com.apple.SpringBoard, com.apple.medialibraryd, com.apple.backboardd, com.apple.mobile.installd, com.apple.SimulatorBridge] to start" UserInfo={NSLocalizedDescription=Timed out waiting for all required services [com.apple.SpringBoard, com.apple.medialibraryd, com.apple.backboardd, com.apple.mobile.installd, com.apple.SimulatorBridge] to start}
```

dig412:
Is this because some of those services don't exist on iOS < 9?

This seems to imply that 'com.apple.backboardd' and 'com.apple.medialibraryd' are missing:
https://github.com/plu/simctl/blob/master/lib/simctl/device.rb#L260

AndreyBelym:
To be precise, `com.apple.medialibraryd` don't run on iOS 8.1, `com.apple.backboardd` is okay.

lawrencelomax:
if that's the case it should be pretty easy to adjust the list of required services to verify a sim is booted, based on the OS Version. This should be a simple PR, if anyone wants to put one up.

lawrencelomax:
The underlying issue looks to be fixed with the PR. Considering this closed. Reopen if there are any further issues.

PR #371, opened 2017-02-03
Don't wait for 'medialibraryd' when booting iOS 8 devices

dig412:
Fixes issue #368.

I've added a test that checks iOS 8 can be booted successfully, but I'm not sure how you ensure which iOS versions are available in the test environment.

### DIFF
--- a/FBSimulatorControl/Strategies/FBSimulatorBootStrategy.m
+++ b/FBSimulatorControl/Strategies/FBSimulatorBootStrategy.m
@@ -421,8 +421,9 @@
 - (NSArray<NSString *> *)requiredLaunchdServicesToVerifyBooted
 {
   FBControlCoreProductFamily family = self.simulator.productFamily;
+  id<FBControlCoreGlobalConfiguration_OS> version = self.simulator.osConfiguration;
   if (family == FBControlCoreProductFamilyiPhone || family == FBControlCoreProductFamilyiPad) {
-    if (FBControlCoreGlobalConfiguration.isXcode8OrGreater) {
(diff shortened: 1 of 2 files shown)


FBSimulatorControl/Strategies/FBSimulatorBootStrategy.m
```diff
@@ -421,8 +421,9 @@ - (FBProcessInfo *)launchdSimPresentWithAllRequiredServices:(NSError **)error
 - (NSArray<NSString *> *)requiredLaunchdServicesToVerifyBooted
 {
   FBControlCoreProductFamily family = self.simulator.productFamily;
+  id<FBControlCoreGlobalConfiguration_OS> version = self.simulator.osConfiguration;
   if (family == FBControlCoreProductFamilyiPhone || family == FBControlCoreProductFamilyiPad) {
-    if (FBControlCoreGlobalConfiguration.isXcode8OrGreater) {
+    if (FBControlCoreGlobalConfiguration.isXcode8OrGreater && [[version versionNumber] doubleValue] >= 9.0) {
```
lawrencelomax: The check itself looks fine, but you can avoid doing this via a `double` and instead use `NSDecimalNumber`, avoiding any hairyness with floating-point numbers.

```[version.versionNumber isGreaterThanOrEqualTo:[NSDecimalNumber numberWithString:@"9.0"]]```

lawrencelomax:
This was implemented in 5c079bb93e2526379d05eaf65af7382f3594fc2c.

PR #340, opened 2016-11-23
Support loading xctestrun files

plu:
This adds a new class `FBXCTestRun` which can load the `.xctestrun` files generated by `xcodebuild build-for-testing`. The benefit of these files are:

* Contains environment variables and arguments, even with expanded variables like `SRCROOT`.
* Populated with tests to skip from Xcode scheme that was used during `build-for-testing`.
* Can be used to populate everything necessary in `FBTestLaunchConfiguration`.

PR #379, opened 2017-02-16
Support testsToRun and testsToSkip

plu:
This adds support for selecting tests that should be run. It's possible
to either focus (testsToRun) on some tests, or to exclude (testsToSkip)
some tests.

Format: "className/methodName"

### DIFF
--- a/XCTestBootstrap/Configuration/FBTestLaunchConfiguration.m
+++ b/XCTestBootstrap/Configuration/FBTestLaunchConfiguration.m
@@ -15,7 +15,7 @@
 
 @implementation FBTestLaunchConfiguration
 
-- (instancetype)initWithTestBundlePath:(NSString *)testBundlePath applicationLaunchConfiguration:(FBApplicationLaunchConfiguration *)applicationLaunchConfiguration testHostPath:(NSString *)testHostPath timeout:(NSTimeInterval)timeout initializeUITesting:(BOOL)initializeUITesting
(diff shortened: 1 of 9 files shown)


plu:
First part that got extracted from #340.

XCTestBootstrap/Configuration/FBTestLaunchConfiguration.m
```diff
@@ -27,14 +27,16 @@ - (instancetype)initWithTestBundlePath:(NSString *)testBundlePath applicationLau
...
+  _testsToSkip = testsToSkip;
 
   return self;
 }
 
 + (instancetype)configurationWithTestBundlePath:(NSString *)testBundlePath
 {
   NSParameterAssert(testBundlePath);
-  return [[FBTestLaunchConfiguration alloc] initWithTestBundlePath:testBundlePath applicationLaunchConfiguration:nil testHostPath:nil timeout:0 initializeUITesting:NO];
+  return [[FBTestLaunchConfiguration alloc] initWithTestBundlePath:testBundlePath applicationLaunchConfiguration:nil testHostPath:nil timeout:0 initializeUITesting:NO testsToRun:nil testsToSkip:nil];
```
lawrencelomax: This is erroneous as the expectation is that these values are never null. It would make more sense for `testsToSkip` to at least default to the empty set.

plu: We could do that. But then later in `XCTestConfiguration` an empty set for `testsToRun` means since Xcode 8.3 beta "don't run any tests". So if we define `nonnull` here in `FBTestLaunchConfiguration`, we need to react to that later on when composing the `XCTestConfiguration`. Would you prefer that? I'll change it then, just let me know :).

plu: In the original pull request I had it like this btw. You can see it in this commit: https://github.com/plu/FBSimulatorControl/commit/1dab86deb599eef2f17cedb6e1ba1f8955a7040e

plu: Changed default of `testsToSkip` to be an empty set.

PR #383, opened 2017-02-21
Skip testScreenshot if Metal is not supported

plu:


lawrencelomax:
We could eventually do this with python decorators, but this works for now.

PR #385, opened 2017-02-26
Retry deleteDevice operation up to three times

plu:
On Travis the deleteDevice operation sometimes fails:

    Domain=NSCocoaErrorDomain
    Code=513
    “B4D-C0-F-F-E” couldn’t be removed because you don’t
    have permission to access it.

Inside of the device's path there's a device.plist which sometimes
cannot be deleted. Probably some process still has an open file
handle to that file.


plu:
I've asked Travis to check 50 times if the test is still randomly failing after this change (see branch `travis`): https://travis-ci.org/plu/FBSimulatorControl/builds

PR #396, opened 2017-03-19
Force dot decimalseparator on NSNumberformatter

robbertvanginkel:
I was was wasting some time over the following command not working:
```
$ fbsimctl --state=booted set_location "37.77493" "-122.419416"
```

As it turns out my Mac's default locale has a `,` decimal separator, which also applies to reading a number with `NSNumberFormatter`. For a commandline utility this doesn't seem very useful, so force a `.` (dot) decimal separator for values parsed into doubles.


lawrencelomax:
Thanks @robbertvanginkel! This implementation looks fine. Could you please add a Unit Test in `CommandParsersTests` to prevent this from regressing in future?

Issue #397, opened 2017-03-21
[fbxctest] Streaming JSON results during test run


alanzeino:
At the moment, `fbxctest` only 'streams' the JSON results at the end of the test run. This isn't a dealbreaker; you still get the full JSON report at the end of the test run. However compared to `xctool`, this might be considered a regression. If you are integrating `fbxctest` with Buck, you don't see the live updating counter as buck runs tests:

```
[+] TESTING...14.8s (26 PASS/0 FAIL)
```

Since I think `fbxctest` is still in alpha, I'd love to help implement the streaming JSON reporting, but I need a little direction.

I've been looking at `FBJSONTestReporter.m` and the general pattern seems to be 'store the dictionary into an array, and at the end of the test run iterate over that array, convert each dictionary into JSON, and print each JSON event to stdout'.

Tried to hack this into working by calling `printEvent:` at the end of `storeEvent:`, and removing the fast enumeration in `printReportWithError:` that prints each event, but this isn't quite correct (though it did work). 

While I think I could mess with the calling order and semantics of these functions I wanted to know what way you prefer, so I don't waste time implementing this the wrong way.

lawrencelomax:
I think this could take a few steps:
1) Pass in a `FBFileConsumer` object in the constructor that wraps a `+[FBFileWriter writerWithFileHandle:NSFileHandle.standardOut blocking:YES]`. This will mean that you can unit-test the `FBJSONTestReporter` by injecting a different Writer and asserting based on the output. Currently this class implicitly writes to `stdout`, which makes it hard to test. Implement the file writing in terms of the `FBFileConsumer`.
2) Extract the creation of 'event dictionaries' to class functions that take all the inputs and return a dictionary. It looks like there is some mixing and matching going on in that class, so simply extracting the creation of these dictionaries would be a huge help.
3) Replace usages of `storeEvent:` with `printEvent:`, this will mean they are written as they appear. This makes the assumption that all calls to `FBJSONTestReporter` will be serialized (I'm not sure if this assumption is correct).

You may have to store *some* state inside this component as you are going along since there are messages like `No end-ocunit event was received, the test bundle has likely crashed`, which depends on some internal state. You could also store all the events, but print them as well. Just make sure not to duplicate them otherwise `printReportWithError:` will double-up the reporting.

lawrencelomax:
I've landed the bulk of 2) in the steps here. I have some changes ready for 3) just need to test it a bit more and will likely land next week.

[...]

PR #399, opened 2017-03-23
Make FBJSONTestReporter take a FBFileConsumer–conforming object for file writing

alanzeino:
Addressing #397 in parts. First is to replace the direct writing to `[NSFileHandle fileHandleWithStandardOutput]` with an `FBFileWriter` which does the same, though on a serial queue.


lawrencelomax:
Thanks for doing this incrementally, makes it a lot easier to merge. I'll import now which will run against a bunch of internal tests so this give us plenty of signal on stability.

Issue #395, opened 2017-03-14
FBAMDCreateDeviceList return nil and crash


jasenhuang:
is FBAMDCreateDeviceList()  deprecated?  return nil and cause crash when CFArrayGetCount
```
+ (NSArray<FBAMDevice *> *)allDevices
{
  NSMutableArray<FBAMDevice *> *devices = [NSMutableArray array];
  CFArrayRef array = FBAMDCreateDeviceList();
  for (NSInteger index = 0; index < CFArrayGetCount(array); index++) {
    CFTypeRef value = CFArrayGetValueAtIndex(array, index);
    FBAMDevice *device = [[FBAMDevice alloc] initWithAMDevice:value];
    if (![device cacheAllValues]) {
      continue;
    }
    [devices addObject:device];
  }
  CFRelease(array);
  return [devices copy];
}
```

lawrencelomax:
This appears to be the case when `usbmuxd` is broken/swamped/not running. We do now have better logging of this at least.

PR #409, opened 2017-05-10
Guard against NULL return from FBAMDCreateDeviceList()

jmoody:
### Motivation

Resolves:

* FBAMDCreateDeviceList return nil and crash #395

### Reproduce

There may be other ways to reproduce, but this is reliable.

* Xcode 8.3.2
* Does not matter if there is a physical device attached to the host machine

```
# Terminate usbmuxd
$ sudo launchctl unload -F /System/Library/LaunchDaemons/com.apple.usbmuxd.plist 
```

Run the FBDeviceControlLinkerTests/testContructsDeviceSet test to see the failure.

```
# Restart usbmuxd
$ sudo launchctl load -wF /System/Library/LaunchDaemons/com.apple.usbmuxd.plist
```

Run the FBDeviceControlLinkerTests/testContructsDeviceSet test to see success.


lawrencelomax:
This looks fixed since the refactor to use `FBAMDeviceManager`

PR #426, opened 2017-07-06
DeviceControl: linker tests can be run in any order

jmoody:
### Reproduce

* Xcode 8.3.3
* macOS Sierra
* At least on physical device attached via USB

    Run this test in FBDeviceControlTests/Tests/Integration/FBDeviceControlLinkerTests.m

```
- (void)testReadsFromMobileDevice
{
  NSArray<FBAMDevice *> *devices = [FBAMDevice allDevices];
  XCTAssertNotNil(devices);
}
```

### Expected

An array of the connected devices.

### Found

```
# EXEC_BAD_ACCESS code=1 address=0x0
FBAMDCreateDeviceList(); 

https://github.com/calabash/FBSimulatorControl/blob/master/FBDeviceControl/Management/FBAMDevice.m#L66
```

### Notes

The problem is that this method: `[FBAMDevice loadFBAMDeviceSymbols];` is never called when the test is run in isolation.

If you run _all_ the tests in this file, then this test passes because a previous test calls methods that load the AMDevice symbols.  If you change the name of the test to `testAAA`, then this test will run first and it will fail.

Moving the load methods out of a test and into the initialize method seems like an acceptable solution.

### Test

I cannot run any of the DeviceControl tests on facebook/FBSimulatorControl master because of Swift dylib loading problems.

### DIFF
--- a/FBDeviceControlTests/Tests/Integration/FBDeviceControlLinkerTests.m
+++ b/FBDeviceControlTests/Tests/Integration/FBDeviceControlLinkerTests.m
@@ -27,10 +27,7 @@
   if (!NSProcessInfo.processInfo.environment[FBControlCoreDebugLogging]) {
     setenv(FBControlCoreDebugLogging.UTF8String, "NO", 1);
   }
-}
 
-- (void)testLinksPrivateFrameworks
-{
   [FBDeviceControlFrameworkLoader initializeEssentialFrameworks];
   [FBDeviceControlFrameworkLoader initializeXCodeFrameworks];
 }


lawrencelomax:
> I cannot run any of the DeviceControl tests on facebook/FBSimulatorControl master because of Swift dylib loading problems

This is being looked at, but this is due `fbsimctl` and the private frameworks that `FBDeviceControl` depends on are from different Swift versions. There can't be different versions of Swift linked within the same process at runtime. You may find that bumping the Swift version to be the same as the linked PrivateFrameworks will make the linker issues go away.

lawrencelomax:
The good news is that we've now removed the dependency on all of these private frameworks, so this is redundant now!

PR #450, opened 2017-10-03
Expose 'list_apps' on HTTP Server

NickAb:
Summary: Exposes the 'list_apps' action on the HTTP Server so that it can be used remotely.


lawrencelomax:
(y)

Issue #454, opened 2017-10-11
Segmentation fault: 11 when calling listApps over http


NickAb:
Calling listApps action over http crashes fbsimctl server with `Segmentation fault: 11`.
So far I traced issue to changes in `iOSRunner.swift` in  https://github.com/facebook/FBSimulatorControl/commit/088d5a9db7b536d489a1684b438ad90977c5bc2a#diff-e937a86125ff321f0e78dd3b1fbd2d4c

If #405 applied to any commit before 088d5a9db7b536d489a1684b438ad90977c5bc2a, then it works fine.

NickAb:
Fixed by https://github.com/facebook/FBSimulatorControl/commit/ff1daf5bf23313a79fb8e47c0c868037d777d058

PR #463, opened 2017-10-20
Fix Segmentation Fault in listApps over HTTP

NickAb:
Summary: Calling `listApps` over HTTP was failing with segfaul due to `dataContainer` being defined as `assign` property.


NickAb:
Fixes #454

Note: running `fbsimctl --state=booted list_apps` was not failing with segfault for some reason.

Issue #459, opened 2017-10-17
support creating iphone 8/ios 11.0 simulators


wei-gu:
right now, not supported. Do you have any plan to support?

zeyadsalloum:
fixed

PR #465, opened 2017-10-23
Add 2017 iPhones to supported targets

NickAb:
Summary: Add iPhone2017-A, iPhone-2017-B, and iPhone-2017-C
to list of supported targets to enable their creation.


NickAb:
Note: iPhone 8, iPhone 8 Plus, and iPhone X device types are specified as iPhone2017-A, iPhone-2017-B, and iPhone-2017-C in `xcrun simctl list devicetypes`. So I went with direct approach adding this device types to available targets. 

Probably it can be made more user-friendly by adding a mapping layer to map from more user friendly `iPhone 8` to `iPhone-2017-A`

NickAb:
Fixes #459

PR #472, opened 2018-03-02
Fix Segmentation Fault in listApps over HTTP

NickAb:
Summary: Calling `listApps` over HTTP was failing with segfaul due to `dataContainer` being defined as `assign` property.


NickAb:
Was fixed in ff1daf5bf23313a79fb8e47c0c868037d777d058

PR #474, opened 2018-03-13
Add ability to pass multiple -only args to run specific UI tests.

beefon:
fbxctest now accepts multiple test filter arguments (`-only`), this allows it to run only specific tests rather than running the whole test suite. Unfortunately, `xctest` binary apparently does not support multiple `-XCTest <filter>` arguments, so UI tests with multiple test filters are only supported for now. Passing multiple filters to logic or app tests will fail the run which is tested.

Test plan: CI, added tests


beefon:
Hi @lawrencelomax , would you be able to have a look please?
Travis CI is completely broken - a lot of tests are failing even without my diff. I've attempted to fix build errors in this diff a bit, but tests are failing.

PR #480, opened 2018-04-30
Return correct registration port value in Future-based API

interstateone:
Hi! Most of the HID implementation is still beyond me, but as I work through trying to use it I noticed what looked like a small error made when this code was rewritten to use FBFutures in 21735b26b394d05f9b2da3c5d4f3235450cc5126.


lawrencelomax:
Thanks! Sorry for the delay, now `idb` is out we're re-reviewing the PRs

PR #486, opened 2018-06-18
Add Alternative Diagnostic Report Path for Simulators

Rhapsod:
Summary: The default user's Diagnostic Report Path does not exist when a root user is used.

Reviewed By: lawrencelomax

Differential Revision: D8346297


lawrencelomax:
Looks like someone inside fb landed a similar fix.

Thanks for the heads-up. Closing.

PR #487, opened 2018-07-17
Add example project.

wigl:
[iSimulator](https://github.com/wigl/iSimulator): GUI utility to control the Simulator, manage the app installed on the simulator.


lawrencelomax:
@wigl I'll be happy to add this to the `FBSimulatorControl/README.md`

Whilst we do now encourage people to use `idb`, yours is an interesting case where the Framework may make more sense.

If you still want this in I'll be happy to merge this, provided it's in the location of the old readme

PR #489, opened 2018-08-06
fix for recording video in simulators

sumanc:
When recording video for simulators using the command;
fbsimctl stream --bgra --fps 12 

I was getting garbled frames. I opened a thread in stack exchange as well - https://superuser.com/questions/1345998/converting-bgra-stream-to-mp4?noredirect=1#comment2016263_1345998

After playing with various pixel format transformation and reading several blogs online, this change is working as expected. I tried on iPhone X simulator with the following command;
./fbsimctl stream --bgra --fps 12 - | ffplay -f rawvideo -video_size 1125x2436  -framerate 12 -pixel_format bgra -

Please let me know if this is the right thing to do.

--Suman

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ output/
 
 # Files copied from Xcode
 Fixtures/Source/FBTestRunnerApp/Frameworks
+out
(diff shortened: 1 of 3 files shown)


sumanc:
BTW, I got inspired by this - https://github.com/BradLarson/GPUImage/blob/master/framework/Source/GPUImageStillCamera.m

to come up the change request. If it is the right thing to do, I will add this link as the comment to credit to the author.

PR #498, opened 2018-11-04
Add a subcommand to enable/disable the hardware keyboard

mykola-mokhnach:
WebDriverAgent fails to enter a text into a text field if the on-screen keyboard is not shown on the simulator and throws an error similar to the one described in https://github.com/facebook/WebDriverAgent/issues/665. With the new command it will be possible to ensure the simulator is configured properly and the keyboard is always shown if needed.

I've also fixed some compiler warnings (mostly about `open<>public` modifier usage) that were preventing fbsimctl from building on Xcode10. Please let me know if it is necessary to revert these.


lawrencelomax:
We'd love to have this in `idb` itself. I'll put up a plan for where we're going with `fbsimctl` but needless to say it's going to be deprecated in favour of `idb` in the long term.

PR #500, opened 2018-11-15
fix implicit conversion error

galkahana:
Compiling head now fails. this takes care of the error, and now the build is Bon.


warent:
#502 is fixed by this. Any progress on getting this merged in?

jeanpaulsio:
Fairly certain https://github.com/facebook/FBSimulatorControl/issues/510 is closed by this as well

galkahana:
master already contains this correction. this can be closed.

PR #506, opened 2019-01-29
Fix typo that breaks builds on case sensitive file systems

EdwardDrapkin:
We were having some trouble building fbsimctl on one of our engineer's machines.  We eventually tracked down the issue to the typo in this file throwing a not found error, due to his filesystem being case sensitive.


zeyadsalloum:
already fixed

Issue #510, opened 2019-03-26
fbsimctl build fails in macOS 10.14.3 and Xcode 10.2 and Swift 5.0 (HEAD@aab3dcd)


douo:
```
==> ./build.sh fbsimctl build /usr/local/Cellar/fbsimctl/HEAD-aab3dcd
using target fbsimctl
using command build
using output directory /usr/local/Cellar/fbsimctl/HEAD-aab3dcd
/usr/local/opt/carthage/bin/carthage
/private/tmp/fbsimctl-20190326-37337-196loy0/fbsimctl /private/tmp/fbsimctl-20190326-37337-196loy0
2019-03-26 16:36:02.286 carthage[37622:13426807] NetworkStorageDB:_openDBReadConnections: failed to open read connection to DB @ /Users/tiaolins/Library/Caches/carthage/Cache.db.  Error=14. Cause=unable to open database file
2019-03-26 16:36:02.286 carthage[37622:13426807] CacheRead: unable to open cache files in /Users/tiaolins/Library/Caches/carthage
2019-03-26 16:36:02.961 carthage[37622:13426807] Warning: No Caches directory could be found or created: You don’t have permission to save the file “org.carthage.CarthageKit” in the folder “Caches”.. (Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “org.carthage.CarthageKit” in the folder “Caches”." UserInfo={NSFilePath=/Users/tiaolins/Library/Caches/org.carthage.CarthageKit, NSUnderlyingError=0x7fcf3e6266a0 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}})
*** Checking out GCDWebServer at "3.3.3"
*** Cloning GCDWebServer
*** xcodebuild output can be found in /var/folders/bt/n4k_1jyx483g34zs_wrysqqh0000gn/T/carthage-xcodebuild.OMArNE.log
*** Building scheme "GCDWebServers (Mac)" in GCDWebServer.xcodeproj
/private/tmp/fbsimctl-20190326-37337-196loy0
User defaults from command line:
    IDEDerivedDataPathOverride = /private/tmp/fbsimctl-20190326-37337-196loy0/build

Build settings from command line:
    SDKROOT = macosx10.14

note: Using new build system
note: Planning build
note: Constructing build description
Build system information
[...]

PR #511, opened 2019-03-26
Convert to Swift 4

donly:
This should fix https://github.com/facebook/FBSimulatorControl/issues/510


tayyab-sarwar:
Same issue!
any fix?

lawrencelomax:
This should now be fixed in some recent commits to master @donly, if there's still changes you need, please feel free to re-open.

Thanks!

PR #516, opened 2019-05-01
Added https:// to the www.fbidb.io link

mbogh:
## Motivation

The link to **fbidb.io** in `README.md` is without `https://` so GitHub recognized the link as internal and you would end up on `https://github.com/facebook/idb/blob/master/www.fbidb.io`.


c-ryan747:
Thanks for spotting this, I'll approve this now

mbogh:
Thanks for the nice work 💯

PR #518, opened 2019-05-05
Fix typos in build script usage texts

BalestraPatrick:
## Motivation

I was reading through the build scripts and noticed a typo in the usage text (non-native speaker here, could be completely wrong too 😄).

## Test Plan

| Before  | After |
| ------------- | ------------- |
| before-1 | after-1 |
| before-2 | after-2 |


lawrencelomax:
"It's" is possessive in this case, which makes the `'s` correct!

Thanks!

PR #527, opened 2019-05-15
Add required dependencies to build idb companion

fr0l:
## Motivation

Add required dependencies to build idb companion

## Test Plan

1. Have a fresh macOS with Xcode
2. Try to compile ```idb_companion```


c-ryan747:
Thanks for spotting this 👍

PR #538, opened 2019-07-16
Add support for iOS 13, tvOS 13, watchOS 6 to iOS target configuration

fr0l:
Add support for iOS 13, tvOS 13, watchOS 6 to iOS target configuration
as well add missing const for 12.1


c-ryan747:
Just merged this in, thanks!

PR #555, opened 2019-09-25
Add iOS 13.1 and iPhone 11 devices

fr0l:
Add iOS 13.1, iPhone 11, 11 Pro, 11 Pro Max, Apple Watch 5

Fix Apple Watch 4 architecture to Arm64


c-ryan747:
Approved, thanks for adding these!

Issue #537, opened 2019-07-15
Unable to run fbsimctl list


nico29:
### Description

Hello, 
I've cloned the repository and built `fbsimctl` using `./build.sh fbsimctl build output` as stated in the readme of the project.
Everything went okay.
Then when I try to to run `fbsimctl list` I get the following output:

```
'["list"]' does not match '[fbsimctl print [action]


[action] Action
===============

	accessibility_fetch
OR	[approve]
OR	[boot]
OR	[clear_keychain]
OR	clone
OR	config
OR	contacts_update <directory>
OR	[create]
OR	delete
OR	[diagnose]
OR	erase
OR	focus
OR	install <string> --codesign
OR	keyboard_override
OR	[launch(agent)]
OR	[launch(app)]
OR	[launch_xctest]
OR	list_apps
OR	list_device_sets
OR	[listen]
OR	list
OR	open <url>
OR	record {{ start <file>? | stop }}
OR	[relaunch]
OR	service_info {{ <application> | <bundle-id> }}
OR	set_location <double> <double>
OR	shutdown
OR	stream {{ --h264 | --bgra }} {{ <fps> | --fps <int> }}? {{ - | <file> }}
OR	tap <double> <double>
OR	logtail <string>* --?
OR	terminate {{ <application> | <bundle-id> }}
OR	uninstall {{ <application> | <bundle-id> }}
OR	upload <file>+
OR	[watchdog_override]

	--bgra                   	Output in BGRA format.
	--codesign               	Before installing, sign the bundle and all its frameworks with a certificate from the keychain
	--fps                    	Frames Per Second of Output
	--h264                   	Output in h264 format.


[approve] Action: Approve
=========================

	approve {{ <application> | <bundle-id> }}+


[boot] Action: Boot
===================

	boot {{ --connect-bridge | --direct-launch | --use-nsworkspace | --scale=25 | --scale=50 | --scale=75 | --scale=100 | <locale> | --locale <locale> }}+

	--connect-bridge
	--direct-launch
	--locale
	--scale=100
	--scale=25
	--scale=50
	--scale=75
	--use-nsworkspace


[clear_keychain] Action: Clear Keychain
=======================================

	clear_keychain {{ <application> | <bundle-id> }}?


[create] Action: Create
=======================

	create {{ --all-missing-defaults | {{ <device-name> | <os-version> | <aux> | --aux <directory> }}* }}

	--all-missing-defaults
	--aux


[diagnose] Action: Diagnose
===========================

[...]

Issue #562, opened 2019-10-10
[Request]idb stream command support


NozomiIto:
### Description

I now use `fbsimctl stream --bgra --fps 10 -` to receive the screen cast in real time but would like to switch to idb command line tool.
But I can't find the corresponding command for idb so I really appreciate if idb has such real time streaming command.

PR #576, opened 2019-12-09
Fix to work `fbsimctl list`

hnisiji:
## Motivation
I fixed the fbsimctl bug reported in #537.
I know fbsimctl is no longer supported, but there are some useful features not yet available in idb, such as stream, so I want to fix it.

## Test Plan
- Build and run `fbsimctl list`

### DIFF
--- a/fbsimctl/FBSimulatorControlKit/Swift/Parser.swift
+++ b/fbsimctl/FBSimulatorControlKit/Swift/Parser.swift
@@ -230,7 +230,7 @@ extension Parser {
     }
 
     let sequentialParser = Parser<()>
-      .ofFlag(flag, explanation)
+      .ofFlag(flag, (), explanation)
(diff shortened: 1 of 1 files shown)


c-ryan747:
Great 😃 Thank you!

c-ryan747:
@hnisiji 

> some useful features not yet available in idb, such as stream

What features would you like to see in idb from fbsimctl? Also what do you mean by stream here?

hnisiji:
I meant  `fbsimctl stream` with stream. I have been use `fbsimctl` like #562.
This feature is important to me.

c-ryan747:
Ah okay. We don't currently have plans to support something like that in idb, however it might be possible to add support for this to the `idb video` command. If thats something you'd be interested in doing then feel free.

Issue #587, opened 2020-03-09
install idb from tarball failed


guyingzhao:
### Description

Install fb-idb via tarball failed:

![image](https://user-images.githubusercontent.com/22538408/76180028-c483a680-61f7-11ea-9d47-46be7e68bf15.png)

### Reproduction

```shell
python3 setup.py sdist
pip3 install dist/fb-idb*
```

### Solution

setup.py forgot to add idb.proto as data files and cased the issue above, just add it would get rid of the trouble.

### Additional Information

* `idb` revision: 1.0.10

guyingzhao:
fixed, refer #588

PR #588, opened 2020-03-09
bugfix for sdist installing

guyingzhao:
## Motivation

We have forked a branch for secondary development, we need to distribute fb-idb to our own pypi server, we build only tarball to pypi server, and when client installs from tarball, an error occurs.

## Test Plan

```shell
python3 setup.py sdist
pip3 install dist/fb-idb-0.0.1.tar.gz --user
```

image

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/idb, and link to your PR here.)


guyingzhao:
related to issue #587

Issue #589, opened 2020-03-11
install idb from tarball failed with "Exception: error: ['grpc_tools.protoc'..."


guyingzhao:
### Description and Reproduction

we have a private pypi server and distribute fb-idb as tarball  fb-idb-*.tar.gz.

install idb from tarball in a newly setup macOS with:

```shell
pip3 install fb-idb -i http://xxx.xxx.xxx.xxx/simple --trusted-host xxx.xxx.xxx.xxx
```
error message:

```shell
    protoc-gen-python_grpc: program not found or is not executable
    Please specify a program using absolute path or make sure the program is available in your PATH system variable
    --python_grpc_out: protoc-gen-python_grpc: Plugin failed with status code 1.
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/81/h_vxlgmx50s9t8yfgw394twm0000gp/T/pip-install-zz01ecw4/fb-idb/setup.py", line 77, in <module>
        cmdclass={"build_py": BuildPyCommand},
      File "/Users/zuosiyang/Library/Python/3.7/lib/python/site-packages/setuptools/__init__.py", line 144, in setup
        return distutils.core.setup(**attrs)
      File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/core.py", line 148, in setup
        dist.run_commands()
      File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/dist.py", line 966, in run_commands
        self.run_command(cmd)
      File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/dist.py", line 985, in run_command
        cmd_obj.run()
      File "/Users/zuosiyang/Library/Python/3.7/lib/python/site-packages/setuptools/command/install.py", line 61, in run
        return orig.install.run(self)
      File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/command/install.py", line 545, in run
        self.run_command('build')
      File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/cmd.py", line 313, in run_command
        self.distribution.run_command(command)
      File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/dist.py", line 985, in run_command
        cmd_obj.run()
      File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/command/build.py", line 135, in run
[...]

PR #591, opened 2020-03-11
compiling grpc proto without depending on grpc binary

guyingzhao:
## Motivation

We have a private pypi server and distribute fb-idb as tarball instead of wheel. And then. we encountered a `grpctools.protoc Error`, I fixed the problem.

More detail can be found in #589 

### Reproduce

* cd to fb-idb root directory and execute `git reset df76dd74523017d8c043a4e5359ec91ae3e3115a`
* `rm /usr/local/bin/protoc-gen-python_grpc` to imitate user environment.
* `python3 setup.py install`

image

### Fixed

* `git pull` and update to modified code
* `python3 setup.py install`

image

## Related PRs

refer issue #589

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,11 +4,10 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-import os
-import setuptools
-import setuptools.command.build_py
-
 from pathlib import Path
+import os
+
+import setuptools.command.build_py
 
 
 class BuildPyCommand(setuptools.command.build_py.build_py):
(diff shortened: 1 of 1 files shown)


c-ryan747:
I think we need to keep the protoc command as it is. The reason we use this protoc plugin is so that we can have the pure python implementation provided by grpclib https://github.com/vmagamedov/grpclib#protoc-plugin

Happy to add the right dependency so that we can get this build working for you though

guyingzhao:
@c-ryan747 I don't think there is a difference between our goal. I'll explain my modification more clearly:

* our user won't git clone idb repository, thus won't have a `protoc-gen-python_grpc` locally
* if we use current compiling command, our user will get an error on `protoc-gen-python_grpc not found`
* if we use the new compiling command, idb still remains in pure python implementation and we don't need the `protoc-gen-python_grpc` binary any more.

PR #592, opened 2020-03-11
dynamiclly generate protoc compiler to avoid installing error for tarball

guyingzhao:
## Motivation

This pull request mainly fixes issue #589 . To ensure the generated proto file to be pure python(see #591 ), I add some code to dynamically generate the expected compiler `protoc-gen-python_grpc` at the root directory, and add root directory to PATH variable.

## Test Plan

1. `rm /usr/local/bin/protoc-gen-python_grpc` since user environment has no such file.

2. disable the `gen_protoc_compiler()` invocation in `setup.BuildPyCommand.run`

3. `python3 setup.py install` and we'll get an error complaining about `protoc-gen-python_grpc` not found

4. enable the `gen_protoc_compiler()` invocation in `setup.BuildPyCommand.run`

5. `python3 setup.py install` will end in success.

### DIFF
--- a/protoc_compiler_template.py
+++ b/protoc_compiler_template.py
@@ -1,0 +1,11 @@
+# EASY-INSTALL-ENTRY-SCRIPT: 'grpclib==0.3.1','console_scripts','protoc-gen-python_grpc'
+__requires__ = 'grpclib==0.3.1'
+import re
+import sys
+from pkg_resources import load_entry_point
+
+if __name__ == '__main__':
+    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
+    sys.exit(
+        load_entry_point('grpclib==0.3.1', 'console_scripts', 'protoc-gen-python_grpc')()
+    )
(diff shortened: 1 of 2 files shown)


c-ryan747:
The code here looks okay, but I'm curious as to why we need to do this. I can uninstall grpclib and still successfully build idb:
```
$ pip3.7 uninstall grpcio-tools grpclib
Skipping grpcio-tools as it is not installed.
Cannot remove entries from nonexistent file /Users/callumryan/Desktop/idb/.eggs/easy-install.pth
$ python3.7 setup.py install
....
// Success
```

Could it be that your local build setup is restricted in some way? Maybe installing the pip or adding to the environment is broken?

guyingzhao:
@c-ryan747 Have you ever tried: `rm /usr/local/bin/protoc-gen-python_grpc` before installing fb-idb?

I have got plenty of feedback that install fb-idb from tarball complaining about `protoc-gen-python_grpc` not found

c-ryan747:
Ah yes I do have it installed in other places. I'm happy with accept this then

PR #647, opened 2020-12-22
Fixed use of FBApplicationLaunchMode for FBDevice

Wumpf:
## Motivation

Previously, an Application on a device would be restarted for FBApplicationLaunchModeForegroundIfRunning and not fail if already running for FBApplicationLaunchModeFailIfRunning.

## Test Plan

Tested this manually with an application that calls directly into FBDeviceControl framework.
This application in turn has some automated tests that capture this behavior (this is how I found the issue).
I have not added any tests to idb, but will happily do so if someone can point me to it (I'm still new to the project).

## Related PRs

-


Wumpf:
apologies, used the wrong github account for this one. Gonna recreate with the correct one

AndreasReich:
(Fixed with #648 - with the CLA in place this I reckon this does matter)

PR #648, opened 2020-12-22
Fixed use of FBApplicationLaunchMode for FBDevice

AndreasReich:
## Motivation

Previously, an Application on a device would be restarted for FBApplicationLaunchModeForegroundIfRunning and not fail if already running for FBApplicationLaunchModeFailIfRunning.

## Test Plan

Tested this manually with an application that calls directly into FBDeviceControl framework.
This application in turn has some automated tests that capture this behavior (this is how I found the issue).
I have not added any tests to idb, but will happily do so if someone can point me to it (I'm still new to the project).

## Related PRs

n/a

### DIFF
--- a/FBDeviceControl/Commands/FBDeviceApplicationCommands.m
+++ b/FBDeviceControl/Commands/FBDeviceApplicationCommands.m
@@ -238,6 +238,26 @@ static void TransferCallback(NSDictionary<NSString *, id> *callbackDictionary, i
 
 - (FBFuture<id<FBLaunchedApplication>> *)launchApplication:(FBApplicationLaunchConfiguration *)configuration
 {
+  if (configuration.launchMode == FBApplicationLaunchModeFailIfRunning) {
+    return [[self processIDWithBundleID:configuration.bundleID] onQueue:self.device.asyncQueue chain:^ (FBFuture<NSNumber *>* processIdQueryResult) {
+      if (processIdQueryResult.state == FBFutureStateDone) {
+        return [[FBDeviceControlError
+          describeFormat:@"Application %@ already running with pid %@", configuration.bundleID, processIdQueryResult.result]
+          failFuture];
+      } else if (processIdQueryResult.state == FBFutureStateFailed) {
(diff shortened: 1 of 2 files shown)


FBDeviceControl/Commands/FBDeviceApplicationCommands.m
```diff
@@ -237,6 +237,26 @@ - (instancetype)initWithDevice:(FBDevice *)device
...
+    return [[self processIDWithBundleID:configuration.bundleID] onQueue:self.device.asyncQueue chain:^ (FBFuture<NSNumber *>* processIdQueryResult) {
+      if (processIdQueryResult.state == FBFutureStateDone) {
+        return [[FBDeviceControlError
+          describeFormat:@"Application %@ already running with pid %@", configuration.bundleID, processIdQueryResult.result]
+          failFuture];
+      } else if (processIdQueryResult.state == FBFutureStateFailed) {
+        return (FBFuture*)[self launchApplicationIgnoreCurrentState:configuration];
+      } else {
+        return (FBFuture*)processIdQueryResult;
+      }
```
c-ryan747, line 251: The final branch herer looks a little suspicious. If the app is running then we fail, but in all other conditions we should probably attempt to launch the app

c-ryan747:
@AndreasReich thanks for working on idb! You'll need to accept the Facebook CLA before we can ship this PR

AndreasReich:
Contacted my legal department already on helping me to fill out the company version of the CLA check before the holidays, just contacted them again :)

FBDeviceControl/Commands/FBDeviceApplicationCommands.m
AndreasReich: You're probably right but maybe we should still not launch it if the future state is cancelled?

c-ryan747: Actually you might be right and we should just propagate the cancel. @lawrencelomax I'm guessing cancelling will be so rare that either is fine

AndreasReich:
@c-ryan747 it's been quite a while, but the CLA should be sorted out by now
Rebased to latest master

PR #758, opened 2022-01-21
WIP: Add tapping of accessibility element using AXLabel

jhildensperger:
## Motivation

Tapping (x,y) is great, but we might not always know the coordinates of an element on all devices. 

In addition to tapping a coordinate, we can leverage the accessibility elements of the screen to find the coordinates of the element we'd like to tap. This is very similar to how developers would write their UITests in their normal workflow.

### Elements supported for interaction
```python
class HIDElementType(Enum):
    back = "back button"
    button = "button"
    text = "text"
    textfield = "text field"
```

### Examples

Tap the first back button on the screen:
`idb ui axtap back`

Tap a button:
`idb ui axtap button "My Button"`

Tap some text:
`idb ui axtap text "some text with an attached gesture recognizer"`

Tap a text field:
`idb ui axtap textfield TextField`

## Test Plan

TBD


idb/grpc/client.py
```diff
@@ -789,6 +791,48 @@ async def list_xctests(self) -> List[InstalledTestInfo]:
...
+                elementFound = axElement == element and (
+                    label is None or axLabel == label
+                )
+
+                if elementFound:
+                    axframe = item["frame"]
+                    x += axframe["x"]
+                    y += axframe["y"]
+                    break
+            except:
```
fgasperij, line 827: What could make it `raise`?

idb/grpc/client.py
```diff
@@ -789,6 +791,48 @@ async def list_xctests(self) -> List[InstalledTestInfo]:
...
+        )
+
+        elementFound = False
+
+        for item in json.loads(accessibilityInfo.json):
+            try:
+                axElement = HIDElementType(item["role_description"])
+                axLabel = item["AXLabel"]
+
+                elementFound = axElement == element and (
```
fgasperij, line 818: For all the element types except for `back`, does it make sense to allow commands without a `label`?

The current behavior is that:

```
$ idb ui axtap button
```

Succeeds and returns the first `button` it encounters, right?

idb/common/types.py
```diff
@@ -121,6 +121,13 @@ class HIDButtonType(Enum):
     SIRI = 5
 
 
+class HIDElementType(Enum):
```
fgasperij, line 124: Could you show an example of each of these from an actual output?

idb/grpc/client.py
```diff
@@ -789,6 +791,48 @@ async def list_xctests(self) -> List[InstalledTestInfo]:
...
+                    y += axframe["y"]
+                    break
+            except:
+                pass
+
+        if elementFound:
+            for n in range(count):
+                await self.send_events(tap_to_events(x, y, duration))
+        else:
+            print("AXElement with AXLabel: ", label, " type: ", element, " not found.")
```
fgasperij, line 834: We could log it, however let's not send it to `stdout` and preserve the current `tap` behavior.

idb/grpc/client.py
jhildensperger: Yes, it just uses the first element of the type.

It's debatable where or not that is useful because I don't know if we can guarantee that the order of `accessibilityInfo.json` is the same every run, though it appeared that way when testing.

idb/grpc/client.py
jhildensperger: I think the enum will `raise` a `ValueError` if it doesn't exist for the role:
`HIDElementType(item["role_description"])`

I'm happy to change that if there is a preferred way of handling this scenario.

idb/grpc/client.py
jhildensperger: 👍 Can change to logging instead of print.

jhildensperger: I would not preserve the current `tap` behavior because the coordinates provided are suppose to be relative to the accessibility element's frame and default to (1,1). It is probably better to just do nothing than send `tap` to unintended coordinates.

jhildensperger:
@fgasperij Thanks for the feedback. I would like to also have some test coverage, but I can't see to get all of the tests to run. Could to please share the command to run the tests?

PR #832, opened 2023-04-29
fix: typos in README.md

saheem128:
## Motivation

I needed an idb installation for my Flipper client. So, I gave it a bit of read and found these small issues.


fejese (approved):
Thanks for the fix

Issue #804, opened 2022-11-06
idb installation of app - best match target version 0.0 error


mattadamson:
Hi

Has anyone seen this error before or know the root cause please?

The best match /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/DeviceSupport/11.0/DeveloperDiskImage.dmg: 11.0 is not suitable for 0.0

I'm using an iphone 6s iOS 15.7 device to launch a test application.  A few other devices do appear to work though which are nearly identical ie the model and iOS version.  

From

https://github.com/facebook/idb/blob/31b6ed0a8c831043601342f0ecbd9ce7b0858594/FBControlCore/Utility/FBDeveloperDiskImage.m

We see this message output

 return [[FBControlCoreError
    describeFormat:@"The best match %@ is not suitable for %ld.%ld", best, targetVersion.majorVersion, targetVersion.minorVersion]
    fail:error];

Because the target version appears to be read as 0.0.  I can't see where / how it's reading this as 0.0

Many thanks

james-3141:
Yes, we are plagued by this!  It is often caused by not having the correct disk image due to not having xCode up to date, but we also see this when we have a large number of physical devices connected.  Did you solve your problem?

james-3141:
I think we may have made progress on this issue, if it is indeed the same one.  In FBDevice.m:

  if (!_name || overwrite) {
    _name = targetInfo.name;
  }
  if (!_osVersion || overwrite) {
    _osVersion = targetInfo.osVersion;
  }
  **_if (\_productVersion || overwrite) {_**
    _productVersion = targetInfo.productVersion;
  }
  if (!_targetType || overwrite) {
    _targetType = targetInfo.targetType;
  }
  if (!_udid || overwrite) {
    _udid = targetInfo.udid;
  }



Here _productVersion always gets set, getting overwritten with "unknown".  A missing '!' ?
Should it be:

  if (**_!_**_productVersion || overwrite) {
    _productVersion = targetInfo.productVersion;
  }

james-3141:
This seems to result in the target version being set to 0.0.0 and then look for the wrong disk image.

PR #839, opened 2023-07-06
To fix "The best match 11.0 is not suitable for 0.0" 

james-3141:
## Motivation

We operate an automated test farm of around 40 iOS devices.  We have struggled with idb trying to load the Developer Disk Image for iOS V0.0 therefore failing with the error text "The best match /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/DeviceSupport/11.0/DeveloperDiskImage.dmg: 11.0 is not suitable for 0.0".

This only seemed to occur when a large number of device are connected to the system (Mac mini M1) with some devices working and others not this being largely dependent on the order that the devices were connected to the host.

This seems to be a simple typo causing the value of _productVersion to be overwritten by all calls to cacheValuesFromInfo.  The value is correctly set initially, but sometimes a later call to the function is made which overwrites it with 'undefined'.  This value is later used to derive the target version to choose the developer disk image.  

I have commented on issue #804 to which this may relate.

## Test Plan

We have a farm of all currently supported iOS devices. Before the fix ~30-50% of the devices would exhibit this issue.  After the fix, none of the devices exhibit this issue. I have attached a picture of a bit of the farm for context.

We have a test script to run through all the devices running our app on each of them and recording any errors. There are none relating to the issue after the fix compared to failures on all test runs prior.

To replicate this:

Connect 16+ different iOS devices to host and loop over all the devices connected running in turn:
idb connect --udid xxxxxxxx-xxxxxxxxxxxxxxx
idb launch my.app.undertest
idb terminate my.app.undertest
idb disconnect

On successful test should be no occurrence of the error : 

"/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/DeviceSupport/11.0/DeveloperDiskImage.dmg: 11.0 is not suitable for 0.0".

We are running a USB 2.0 hub tree with 48 USB connections hosted by a Mac Mini M1.  All devices are different in our test fixture.

The problem starts to occur in our set up occurs if more than ~8 devices are connected but that is not entirely consistent. An individual 16 port hub (internally 5 x 4 port hubs) shows this problem when all ports are used.

IMG_9557

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/idb, and link to your PR here.)

### DIFF
--- a/FBDeviceControl/Management/FBDevice.m
+++ b/FBDeviceControl/Management/FBDevice.m
@@ -199,7 +199,7 @@
   if (!_osVersion || overwrite) {
     _osVersion = targetInfo.osVersion;
   }
-  if (_productVersion || overwrite) {
+  if (!_productVersion || overwrite) {
     _productVersion = targetInfo.productVersion;
   }
   if (!_targetType || overwrite) {


xgerrit:
I'm sorry that PRs weren't being reviewed for a long time. This was a good catch. Thank you for contributing and love the pic of the device farm!

PR #851, opened 2023-10-09
Fix get_default_companion_path for Apple Silicon

heath-clink:
On Apple Silicon systems, homebrew installs into /opt/homebrew instead of /usr/local. Update get_default_companion_path to look for idb_companion in both /opt/homebrew/bin and /usr/local/bin (if it is not already in PATH)

## Motivation

On newer macOS hardware, homebrew installs into /opt/homebrew by default.

1. Most users probably install idb via homebrew
2. /opt/homebrew/bin is not in the default PATH that apps see when launched from Finder
3. So when you launch (e.g.) Flipper from Finder it will log an error like "/usr/local/bin/idb_companion not found" and idb functionality won't work.
4. Therefore, this change makes get_default_companion_path look for idb_companion in /opt/homebrew/bin then /usr/local/bin (if it's not already in PATH) to cover all the likely configurations.

## Test Plan

With this change, Flipper works correctly

### DIFF
--- a/idb/cli/main.py
+++ b/idb/cli/main.py
@@ -118,7 +118,13 @@ logger: logging.Logger = logging.getLogger()
 def get_default_companion_path() -> Optional[str]:
     if sys.platform != "darwin":
         return None
-    return shutil.which("idb_companion") or "/usr/local/bin/idb_companion"
+    for path in (
+        shutil.which("idb_companion"),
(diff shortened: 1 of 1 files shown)


heath-clink:
It's uncertain if my employer will be willing to sign the CLA, and since I implemented this change on company time I don't think I should use a personal account. Perhaps someone else might be willing to take this change and open a new PR.

PR #857, opened 2024-01-09
Add visionOS as a known platform.

ivanhernandez13:
This change allows identifying Apple Reality Pro devices as visionOS devices when working with FBDevice/FBSimulator objects rather than generic unknown products.


xgerrit:
I'm sorry PRs weren't being reviewed for a long time. I tried to merge this, but too many changes have landed and it's no longer mergable. Thanks for submitting

PR #861, opened 2024-02-07
Convert the CircleCI workflow to GitHub Actions workflows

jk0:


fgasperij:
@jk0 Is this working or not? I don't understand if the issue you mention is resolved or not.

jk0:
@fgasperij 👋 I got it working as far as I could without actually pushing changes in the final steps of the workflow. Please feel free to test it on your end after setting the `GITHUB_TOKEN` secret variable and let me know how it goes. If all goes well, we can proceed to merge the changes.

PR #881, opened 2025-06-02
allow specifying work/async queues via configuration

rubu:
## Motivation

I need to run idb in a node native addon thus there is nothing to drive the main queue which is used under the hood. This PR seems to allow me to boot a device and receive the callbacks by providing  custom  async/work queues via `FBSimulatorControlConfiguration`. I tried to make this an opt in behaviour - the new properties are default `nil` and have a designated initializer and internally the work/async queues use the old values if these configuration values are not specified.

## Test Plan

Testing is something i wanted some feedback on. So obviously:

1) using the existing configuration should not change any tests since as i mentioned this should be opt in
2) tests also use `dispatch_get_main_queue()` which would then need to be adjusted - as in the tests should then use a given queue (most likely each test should be run with main and with a custom queue?), provide it to the frameworks via the configuration and expect everything to work. Is that the route that should be taken here?

### DIFF
--- a/FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.h
+++ b/FBSimulatorControl/Configuration/FBSimulatorControlConfiguration.h
@@ -27,6 +27,18 @@ NS_ASSUME_NONNULL_BEGIN
 + (instancetype)configurationWithDeviceSetPath:(nullable NSString *)deviceSetPath logger:(nullable id<FBControlCoreLogger>)logger reporter:(nullable id<FBEventReporter>)reporter;
 
 /**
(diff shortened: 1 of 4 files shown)


xgerrit:
I'm sorry that PRs weren't being reviewed for a significant amount of time. We've since updated a lot of the code to use Swift Concurrency, which may or may not help with the Node issue, but also means the project is moving away from DispatchQueues entirely. Thank you for contributing this though.

rubu:
All good, i've worked around it, but https://github.com/facebook/idb/pull/885 might be valuable.

PR #885, opened 2025-06-30
implement device orientation

rubu:
## Motivation

`idb` currently does not support changing device rotation (equivalent of "rotate left" and "rotate right" functionality of Simulator.app)

## Test Plan

This PR currently does not have support in the python client, once that is done this could be tested by sending the command and then comparing the resulting screen. I'm not currently sure if there is a reliable back channel for determining when the rotation has finished.


rubu:
I see how much things have diverged since this was opened, will see if i can rebase this on current main.

rubu:
Found that this has been added min main, closing this.

PR #888, opened 2025-08-22
Fix up bad param documentation

dmaclach:
`queue` is not a parameter for this method.

## Motivation

This broke when I was attempting to build


xgerrit:
I'm sorry there was nobody reviewing PRs for a while. It looks like this was fixed at some point. Thanks for submitting this though!

PR #901, opened 2026-02-12
Fix describe remote web content discovery and add regression tests

lapfelix:
## Motivation

`idb ui describe` was not reliably returning accessibility elements inside embedded web content (WebView), even when those elements were present and VoiceOver-accessible.

Root causes:
- The `describe` path did not enable remote-content discovery options.
- Remote discovery could be starved when native traversal reported near-total frame coverage (common with full-screen proxy/container elements), because covered points were skipped and same/frontmost PID hits were filtered out.
- Coverage accounting included `AXWeb*` container roles, which could overstate “already covered” area and suppress remote probing.

Implemented changes:
- `CompanionLib/FBIDBCommandExecutor.m`: enabled remote discovery for describe calls by setting `collectFrameCoverage` and default `remoteContentOptions`.
- `FBSimulatorControl/Commands/FBSimulatorAccessibilityCommands.m`: excluded web container roles (`AXWeb*` / `Web*`) from base frame coverage; added near-full-coverage fallback probing (`coverage >= 0.98`); allowed seen/frontmost/unknown PID hits in that fallback path to recover hidden content; skipped near-fullscreen proxy hits to avoid duplicate wrappers; kept explicit unsigned index conversions needed for strict warning-as-error builds.
- `FBSimulatorControlTests/Utilities/AccessibilityDoubles.h` and `FBSimulatorControlTests/Utilities/AccessibilityDoubles.m`: added translation-to-element mapping support in the translator double so tests can model frontmost-tree vs hit-tested remote elements.
- `FBSimulatorControlTests/Tests/Unit/FBSimulatorAccessibilityCommandsTests.m`: added targeted regression tests:
  - `testCoverageCalculationSkipsWebContentRoleElements`
  - `testRemoteDiscoveryFindsHiddenContentWhenCoverageIsNearFull`
  - `testRemoteDiscoverySkipsFullscreenProxyInNearFullCoverageMode`
- `FBSimulatorControlTests/Tests/Unit/FBSimulatorAccessibilityCommandsTests.m`: retained test-only `int64_t` casts needed to satisfy strict sign-compare checks in this toolchain.

Compatibility:
- No gRPC interface changes.
- No CLI schema/flag changes.
- Behavior changes are scoped to accessibility describe discovery and supporting tests.

## Test Plan

1. Build tests target:
   ```bash
   xcodebuild -project FBSimulatorControl.xcodeproj -target FBSimulatorControlTests -configuration Debug -sdk macosx build
   ```

2. Run accessibility unit suite:
   ```bash
   export DYLD_FRAMEWORK_PATH=$PWD/build/Debug
   xcrun xctest -XCTest FBSimulatorControlTestCase,FBSimulatorAccessibilityCommandsTests $PWD/build/Debug/FBSimulatorControlTests.xctest
   ```
   Result: `21 tests, 0 failures` (includes all new regression tests).

3. Build companion libraries:
   ```bash
   xcodebuild -project FBSimulatorControl.xcodeproj -target CompanionLib -configuration Debug -sdk macosx build
   ```

4. Manual runtime verification on booted simulator:
   ```bash
   build/Build/Products/Debug/idb_companion --udid 78F1F8EA-CDE6-45A6-BB20-C5D40241D8E6 --grpc-port 11601 --log-level debug
   idb --companion 127.0.0.1:11601 ui describe-all --json
   ```
   Verified output includes embedded webview labels/element.

## Related PRs

None.

I accepted the CLA

### DIFF
--- a/CompanionLib/FBIDBCommandExecutor.m
+++ b/CompanionLib/FBIDBCommandExecutor.m
@@ -155,6 +155,8 @@ FBFileContainerKind const FBFileContainerKindFramework = @"framework";
       FBAccessibilityRequestOptions *options = [FBAccessibilityRequestOptions defaultOptions];
       options.nestedFormat = nestedFormat;
       options.enableLogging = YES;
+      options.collectFrameCoverage = YES;
+      options.remoteContentOptions = [FBAccessibilityRemoteContentOptions defaultOptions];
 
       if (value) {
         return [commands accessibilityElementAtPoint:value.pointValue options:options];
--- a/FBSimulatorControl/Commands/FBSimulatorAccessibilityCommands.m
+++ b/FBSimulatorControl/Commands/FBSimulatorAccessibilityCommands.m
@@ -313,8 +313,10 @@ static const CGFloat kDefaultCellSize = 10.0;
 
   // Fill cells row by row using memset for efficiency
   NSUInteger fillWidth = (NSUInteger)(maxX - minX + 1);
+  NSUInteger startX = (NSUInteger)minX;
   for (NSInteger y = minY; y <= maxY; y++) {
-    memset(&_grid[y * _width + minX], 1, fillWidth);
+    NSUInteger row = (NSUInteger)y;
+    memset(&_grid[row * _width + startX], 1, fillWidth);
   }
 }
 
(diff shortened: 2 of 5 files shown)


kzlar:
@lapfelix This is exactly what I've been looking for! Appreciate the work on this, it works well for me. Really hope it could be merged into the main branch.

Have you looked into interacting with the "Sign in with Apple" modals? idb seems pretty unreliable in describing them and I'm wondering if it could be improved in a similar way?

lawrencelomax:
I think a lot has changed since this PR was filed (sorry for the delay).

Quite happy to have another look at this, but recent changes should mean that this is viable using the `axbridge` without the earlier "point grid" hack for the old `ax` backend.

If this doesn't work for you then lmk. At worst case we can expose the "point grid" approach in the old backend. However, the performance here won't be great (you're talking 100s of ms due to XPC overheads).

lapfelix:
Yeah agreed. I'll come back to this when I need it. Will close for now

PR #923, opened 2026-05-16
refactor: Modernize Task.select to use structured TaskGroup

inju2403:
##  Motivation

  While exploring the Swift side of idb, I noticed that Task.select in IDBCompanionUtilities is implemented using withUnsafeContinuation combined with an Atomic-backed state machine — a pattern that
  predates Swift's structured concurrency primitives. withTaskGroup is the standard-library primitive for this exact "first-to-complete" race, so this seemed like a natural modernization.

##  What changes

  - Task.select is reimplemented on top of withTaskGroup.
  - The internal TaskSelectState helper is removed.
  - The public API (both the Sequence overload and the variadic overload) is unchanged.
  - An empty-input precondition replaces the prior silent hang.
  - Docstrings are expanded to clarify the cancellation semantics that the previous implementation also exhibited.

##  Behavior preservation

  The previous implementation had two specific behaviors that callers depend on, both of which are preserved:

  1. Losing tasks are not auto-cancelled. Callers retain ownership of the remaining input tasks and choose when to cancel them (e.g., VideoStreamMethodHandler cancels the stream explicitly after
  Task.select returns).
  2. Outer cancellation cancels all input tasks via the withTaskCancellationHandler onCancel: block.

  Existing call sites in VideoStreamMethodHandler, LogMethodHandler, and TaskTimeout require no changes.

##  Test Plan

  The three existing Task.select tests pass unchanged, validating behavioral equivalence with the previous implementation:

  - testSelectReturnsFirstCompletedTask
  - testSelectWithSequence
  - testSelectCancellationCancelsTasks

  Three new tests are added to lock in previously-untested behavior:

  - testSelectReturnsFailingTaskWhenItFinishesFirst — verifies a task that throws first is correctly returned as the winner.
  - testSelectWithSingleTaskReturnsThatTask — boundary coverage.
  - testSelectLeavesLosersRunning — explicitly asserts that losing tasks continue running after a winner is selected, codifying the previously-undocumented contract.

  Indirect coverage via TaskTimeout (which uses Task.select internally) — the existing testTimeoutSucceedsWhenJobCompletesInTime, testTimeoutThrowsWhenJobExceedsTimeout, and testTimeoutPropagatesJobError
  exercise the new implementation end-to-end.

##  Related PRs

  N/A — this is a standalone refactor with no API surface change.

### DIFF
--- a/IDBCompanionUtilities/TaskSelect.swift
+++ b/IDBCompanionUtilities/TaskSelect.swift
@@ -7,71 +7,62 @@
 
 import Foundation
 
-struct TaskSelectState<Success: Sendable, Failure: Error>: Sendable {
-  var complete = false
-  var tasks: [Task<Success, Failure>]? = []
+extension Task where Success: Sendable {
 
-  mutating func add(_ task: Task<Success, Failure>) -> Task<Success, Failure>? {
-    if var tasks {
-      tasks.append(task)
-      self.tasks = tasks
-      return nil
-    } else {
-      return task
-    }
-  }
-}
-
-extension Task {
-  /// Determine the first task to complete of a sequence of tasks.
+  /// Returns the first task to complete out of a sequence of tasks.
   ///
-  /// - Parameters:
-  ///   - tasks: The running tasks to obtain a result from
-  /// - Returns: The first task to complete from the running tasks
+  /// Once a winner is determined, the remaining tasks are left to finish on
+  /// their own (matching prior behavior). If the surrounding task is
+  /// cancelled before a winner is selected, cancellation is propagated to
+  /// every input task.
+  ///
+  /// - Parameter tasks: The running tasks to await.
+  /// - Returns: The first task to complete, whether by success or failure.
+  /// - Precondition: `tasks` must not be empty.
   public static func select<Tasks: Sequence & Sendable>(
     _ tasks: Tasks
   ) async -> Task<Success, Failure>
-  where Tasks.Element == Task<Success, Failure>, Success: Sendable {
+  where Tasks.Element == Task<Success, Failure> {
 
-    let state = Atomic<TaskSelectState<Success, Failure>>(wrappedValue: .init())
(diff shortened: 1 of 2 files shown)


xgerrit:
Sorry for the delay on this. When it went through internal code review the following issue was raised:

"The new implementation may not preserve the original "first-to-complete return" semantics of Task.select. When a withTaskGroup body returns, Swift's structured concurrency implicitly awaits the remaining child tasks rather than cancelling them. Each child task here does _ = await task.result, so it only finishes when its corresponding input task finishes. The net effect is that withTaskGroup (and therefore Task.select) will not actually return until every input task has completed — not just the first one. This contradicts both the in-code comment ("Leaving the group implicitly cancels the remaining observer tasks via structured concurrency") and the docstring ("the remaining tasks are left to finish on their own"), and it changes the observable behavior versus the previous withUnsafeContinuation-based implementation, which resumed as soon as any task completed."

inju2403:
Thanks @xgerrit, you're right, and I confirmed it.

When a withTaskGroup body returns, the group awaits its remaining child tasks instead of cancelling them. Each child here just does await task.result on a caller-owned task, so none of them finish until their underlying task does. That means select blocks until every input completes, not just the first. I reproduced it: with one task finishing in ~0.1s and two in ~2s, select returned after ~2.0s instead of ~0.1s.

group.cancelAll() doesn't help either. Task.result isn't interrupted by cancellation, so the children still wouldn't return early, and cancelling them would break this API's contract that losing tasks keep running under the caller's ownership.

The root cause is that this contract (return on first completion, leave the others running un-cancelled) is inherently unstructured: the losing tasks have to outlive the select scope, which is exactly what a structured withTaskGroup prevents. The original withUnsafeContinuation implementation is the right tool here.

If it's useful, I'd be happy to open a smaller follow-up that keeps the current implementation and only adds the empty input precondition, expanded docs, and the new tests. Thanks for the careful review, learned a lot from this one.

PR #947, opened 2026-07-28
Select DTUHID from sandboxed clients

AvdLee:
## Summary
- probe the per-simulator DTUHID bootstrap service instead of enumerating host processes
- ensure sandboxed consumers choose DTUHID when modern CoreSimulator suppresses Indigo
- publish regenerated FBSimulatorControl artifacts with Xcode 26.6 provenance

## Test plan
- [x] Focused `FBSimulatorDTUHIDTransportTests`
- [x] Regenerate xcframeworks with Xcode 26.6
- [x] Runtime linkage verification
- [ ] Validate tap, swipe, long press, text input, multitouch, and hardware buttons from sandboxed RocketSim


AvdLee:
Opened against upstream by mistake; this fork-specific binary artifact change belongs in AvdLee/idb.

PR #953, opened 2026-08-10
Load SimulatorKit before the Indigo HID client lookup to avoid clientClassUnavailable

### LINKED ISSUE #941: Indigo HID transport constructs the HID client before SimulatorKit is loaded — clientClassUnavailable for library consumers
## Summary

`FBSimulatorIndigoHIDTransport.indigo(for:)` evaluates its initializer arguments in declaration order:

```swift
static func indigo(for simulator: FBSimulator) throws -> FBSimulatorIndigoHIDTransport {
  FBSimulatorIndigoHIDTransport(
    indigoClient: try FBSimulatorIndigoHIDClient(for: simulator.device),  // 1st: objc_lookUpClass("SimulatorKit.SimDeviceLegacyHIDClient")
    indigo: try FBSimulatorIndigoHID(),                                   // 2nd: loads SimulatorKit via xcodeFrameworks loader
    ...
```

`FBSimulatorIndigoHIDClient.init(for:)` looks up `SimulatorKit.SimDeviceLegacyHIDClient` with `objc_lookUpClass`, but the thing that actually loads SimulatorKit into the process — `FBSimulatorControlFrameworkLoader.xcodeFrameworks.loadPrivateFrameworks()` inside `FBSimulatorIndigoHID.init()` — runs **after** it. Any process that has not already loaded SimulatorKit by other means throws `FBSimulatorHIDError.clientClassUnavailable` on the entire Indigo path.

This includes the **default transport selection**: on a simulator where `dtuhidd` is not active, `FBSimulator.defaultHIDTransport` resolves to `.indigo`, so a plain `FBSimulatorHID(for: simulator)` fails.

## Why idb_companion does not hit this

`FBIDBCommandExecutor.connectToHID()` calls `FBSimulatorControlFrameworkLoader.xcodeFrameworks.loadPrivateFrameworks(...)` before `connectToHID`, so inside the companion SimulatorKit is always resident before the client lookup runs. The bug only surfaces for consumers that link `FBSimulatorControl` directly and follow the documented HID entry point.

## Reproduction

main @ c51004c9, macOS 26, Xcode 27.0 beta 4 (`DEVELOPER_DIR` pointing at it), CoreSimulator 1169.1, booted iOS 27.0 simulator, Device Hub closed (no `dtuhidd`, so the indigo path is selected):

```swift
try FBSimulatorControlFrameworkLoader.essentialFrameworks.loadPrivateFrameworks(logger)
let config = FBSimulatorControlConfiguration(deviceSetPath: nil, logger: logger, reporter: nil)
let set = try FBSimulatorControl.withConfiguration(config).set
let simulator = set.simulator(withUDID: udid)!
let hid = try FBSimulatorHID(for: simulator)   // throws clientClassUnavailable
```

With `FBCONTROLCORE_DEBUG_LOGGING=1` the log confirms only CoreSimulator was ever loaded:

```
CoreSimulator: Already loaded, skipping
Loaded All Private Frameworks [CoreSimulator]
FAIL: clientClassUnavailable(className: "SimulatorKit.SimDeviceLegacyHIDClient")
```

Forcing `transport: .dtuhid` works fine in the same process (no SimulatorKit needed), which also confirms the rest of the stack is healthy.

## Suggested fix

Load the Xcode frameworks (or at least SimulatorKit) at the top of `indigo(for:)` before constructing `FBSimulatorIndigoHIDClient`, e.g.:

```swift
static func indigo(for simulator: FBSimulator) throws -> FBSimulatorIndigoHIDTransport {
  try FBSimulatorControlFrameworkLoader.xcodeFrameworks.loadPrivateFrameworks(nil)
  ...
```

or reorder so `FBSimulatorIndigoHID()` is constructed first. Consumers can work around it today by pre-loading `xcodeFrameworks` themselves before creating an `FBSimulatorHID`.

inju2403:
Fixes #941

## Motivation

As reported in #941, `FBSimulatorIndigoHIDTransport.indigo(for:)` builds its two failable arguments in declaration order, which Swift evaluates left-to-right: `FBSimulatorIndigoHIDClient(for:)` — which looks up the runtime-only SimulatorKit class `SimDeviceLegacyHIDClient` with `objc_lookUpClass` — is evaluated before the sibling `FBSimulatorIndigoHID()`, the only thing on this path that loads SimulatorKit (via `xcodeFrameworks.loadPrivateFrameworks()`). So the lookup runs before SimulatorKit is resident, and any process that hasn't already loaded it fails the entire Indigo path with `clientClassUnavailable`.

`idb_companion` doesn't hit this because `FBIDBCommandExecutor.connectToHID()` pre-loads `xcodeFrameworks` before connecting. It only surfaces for consumers that link `FBSimulatorControl` directly and follow the documented HID entry point — `FBSimulatorHID(for:)`, which on a simulator without an active `dtuhidd` selects the `.indigo` transport, so a plain `FBSimulatorHID(for: simulator)` fails. The sibling `FBSimulatorIndigoHID.init()` already loads SimulatorKit for its own `dlsym` needs; `FBSimulatorIndigoHIDClient` is the one class lookup on this path that relies on someone else having loaded it first.

## Test Plan

The fix loads SimulatorKit at the top of `FBSimulatorIndigoHIDClient.init(for:)`, before the lookup, so the type that needs the class is the one that guarantees the framework is resident — mirroring how the sibling `FBSimulatorIndigoHID.init()` already loads the framework it needs. The loader is `dispatch_once`/`hasLoadedFrameworks`-guarded, so it is a no-op wherever SimulatorKit is already loaded (all companion HID flows), leaving existing behavior unchanged. The fixes suggested in #941 — loading in `indigo(for:)`, or reordering the two arguments — are both sound and resolve the ordering directly; I went with loading inside the client's own initializer so that correctness stays independent of the factory's argument-evaluation order and lives next to the code that depends on it.

The `FBSimulatorControl` framework builds. The mechanism is confirmed by a standalone probe: `objc_lookUpClass("SimulatorKit.SimDeviceLegacyHIDClient")` returns nil before SimulatorKit is loaded (the exact condition that throws `clientClassUnavailable`) and resolves the class once it is `dlopen`-ed. The full end-to-end failure needs a direct `FBSimulatorControl` library consumer on a booted simulator with `dtuhidd` off, so it isn't covered by a unit test here; the fix is verified by inspection, by that probe, by the reproduction in #941, and by parity with the already-working companion preload.

## Related PRs

None.

### DIFF
--- a/FBSimulatorControl/HID/FBSimulatorIndigoHIDClient.swift
+++ b/FBSimulatorControl/HID/FBSimulatorIndigoHIDClient.swift
@@ -50,6 +50,10 @@ final class FBSimulatorIndigoHIDClient: @unchecked Sendable {
 
   /// Looks up, allocates and initializes the runtime-only HID client for the provided device.
   convenience init(for device: SimDevice) throws {
+    // `objc_lookUpClass` returns nil until SimulatorKit is resident, so load it before the lookup
+    // instead of relying on an earlier caller — mirroring the sibling `FBSimulatorIndigoHID.init()`.
+    // The loader is `dispatch_once`-guarded, so this is a no-op when SimulatorKit is already loaded.
+    try FBSimulatorControlFrameworkLoader.xcodeFrameworks.loadPrivateFrameworks(nil)
     guard let clientClass = objc_lookUpClass(Self.clientClassName) else {
       throw FBSimulatorHIDError.clientClassUnavailable(className: Self.clientClassName)
     }


onevcat:
@onevclaw This task is related to issue #941, which we previously submitted. Could you review this PR, report anything you find, and help improve it?

onevclaw:
I reviewed the change against #941 and found no correctness blocker. Loading the framework in `FBSimulatorIndigoHIDClient.init(for:)` makes the `objc_lookUpClass` dependency self-contained, so the fix no longer relies on the factory's argument-evaluation order. The loader only covers SimulatorKit and is already guarded, so the repeated call is appropriately cheap.

The one improvement I would suggest is a regression test that exercises direct `FBSimulatorHID(for:)` after loading only the essential frameworks and verifies it does not fail with `clientClassUnavailable`. If that scenario cannot be made reliable in CI, a documented manual regression run against a booted simulator would still make the validation more durable.

The current CI failures appear unrelated to this four-line change: the macOS test schemes are not configured for testing, the distribution build has an existing header-path failure, the framework build stops in unrelated XCTestBootstrap sendability code, and the Python build lacks `pkg_resources`. None of the reported failures reference this HID client or the new loader call.

> 🧵 mh_1786361845871_poll-one · 🐾 main → main
> 🪝 gh:facebook:idb:issue:953:main:m202608
> 🆔 a0ef2403-a5b0-47ce-8f32-73357d2bf0b5

bryanyantx:
Thanks for putting up this fix! The diagnosis is correct, but it looks like [af6a5d2](https://github.com/facebook/idb/commit/af6a5d2604c97f9c5ac0fca27b9ebe59305b613b) with an equivalent fix has landed on main, so I'm closing this PR as already resolved.

PR #955, opened 2026-08-24
Fix single-AsyncIterator crash in multi-frame gRPC request handlers

erneestoc:
## Summary

`GRPCAsyncRequestStream` is backed by `NIOThrowingAsyncSequenceProducer`, which fatal-errors if more than one `AsyncIterator` is ever created from it:

```
NIOThrowingAsyncSequenceProducer allows only a single AsyncIterator to be created
```

The `AsyncSequence.requiredNext` helper (`Companion/Utility/AsyncSequence+Extension.swift`) reads via `first(where:)`, and **every** `first(where:)` call constructs a fresh iterator; a `for try await … in requestStream` loop makes one too. Any handler that combines two such reads on the same request stream aborts the whole companion on the second read.

Affected RPC handlers (directly, or via the shared `MultisourceFileReader`):

| Handler | Second iterator comes from | Command |
|---|---|---|
| `InstallMethodHandler` | payload frames + trailing `for try await … in requestStream` | `idb install` |
| `RecordMethodHandler` | start, then stop | `idb video` / screen recording |
| `LaunchMethodHandler` | start, then stop | `idb launch --wait-for` |
| `TailMethodHandler` | start, then stop | file tail |
| `XctraceRecordMethodHandler` | start, then stop | `idb xctrace record` |
| `InstrumentsRunMethodHandler` | start, then stop | `idb instruments` |
| `DapMethodHandler` | start, then `consumeElements` loop | `idb dap` |
| `ReplMethodHandler` | start, then `serve` loop | `idb repl` |
| `VideoStreamMethodHandler` | start, then client-cancel loop | `idb video-stream` |
| `PushMethodHandler` / `AddMediaMethodHandler` | `MultisourceFileReader` reads `requiredNext` then loops the rest | `idb file push`, `idb add-media` |

## Fix

Introduce a shared `SingleIteratorRequestStream` (`Companion/Utility/`) that makes **one** iterator per RPC and exposes `next()` / `requiredNext` against it. Each multi-frame handler wraps its `requestStream` once and routes every read through the wrapper; handlers that delegate to a helper (`dap`, `repl`, `push`, `add-media`) thread the wrapper into the helper, and `MultisourceFileReader` now takes the wrapper as well. The trailing `for try await request in requestStream` loops become `while let request = try await stream.next()` on the same iterator.

The `requiredNext` extension is kept for the many genuine single-read handlers, now with a doc-comment warning about the single-shot hazard.

## Notes

- Reproduced with grpc-swift 1.23.1 / current SwiftNIO. The trap is inherent to grpc-swift's single-iterator contract, so this reproduces wherever a recent grpc-swift/NIO is used; please confirm against the internal pin.
- Verified the companion builds (Xcode 26.2 / Swift 6.2) and that `idb video` now records a valid MP4 instead of aborting.
- No behavior change for the single-read handlers (`debugserver`, `hid`, `video-stream`-single-frame paths, etc.).

### DIFF
--- a/Companion/SwiftServer/MethodHandlers/AddMediaMethodHandler.swift
+++ b/Companion/SwiftServer/MethodHandlers/AddMediaMethodHandler.swift
@@ -15,9 +15,13 @@ struct AddMediaMethodHandler {
   let commandExecutor: FBIDBCommandExecutor
 
   func handle(requestStream: GRPCAsyncRequestStream<Idb_AddMediaRequest>, context: GRPCAsyncServerCallContext) async throws -> Idb_AddMediaResponse {
+    // grpc-swift traps if a second AsyncIterator is created; read every
+    // request frame through one owned iterator.
+    let stream = SingleIteratorRequestStream(requestStream)
+
     let extractedFileURLs =
       try await MultisourceFileReader
-      .filePathURLs(from: requestStream, temporaryDirectory: commandExecutor.temporaryDirectory, extractFromSubdir: true)
+      .filePathURLs(from: stream, temporaryDirectory: commandExecutor.temporaryDirectory, extractFromSubdir: true)
 
     try await commandExecutor.add_media(extractedFileURLs)
     return .init()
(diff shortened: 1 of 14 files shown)


erneestoc:
@lawrencelomax been working on [rules_idb for bazel](https://github.com/erneestoc/rules_idb) to allow me to run more than a few concurrent simulators for my test runs (xcodebuild test-without-building is using too much memory for some reason). Let me know if there's something I can help with.

lawrencelomax:
Hiya @erneestoc!

This was due to a difference in the grpc version used between the internal builds and OSS builds that has remained dormant for years. This has since been resolved on master and there's also been grpc server upgrades along the way. Internal and OSS builds are now pinned to use the same grpc version and there are e2e tests running in internal and public CI that should prevent this from happening in future.

Please reopen or refile if this doesn't match your expectations but I think you should find that this is resolved within the last week on master. The latest release should also have the fixes in.

Thanks for reporting!

PR #957, opened 2026-08-31
Continue the single fetch across process boundaries

### LINKED ISSUE #956: Web view content and other out-of-process subtrees are missing from the default accessibility read
## Impact

Since [`27c7821`](https://github.com/facebook/idb/commit/27c7821677b33a51237ae31886066d81daa14ef2) ("Read the guest tree in one fetch by default"), a whole-tree read on the `axbridge` backends does not return anything drawn by another process. The subtree's root node is returned with zero children.

In stock Safari that means **the entire web page**. WebKit renders out of process, so the read returns Safari's chrome and stops at the web view. 

**Important**: This affects more than just WebKit or Safari, it's just the easiest way to demonstrate the issue. The same applies to any app embedding a `WKWebView`,  and to photo pickers, document pickers and iOS's "Use Strong Password?" autofill sheet.

We drive simulators over the companion's gRPC API, as a result:

- **Page and sheet content cannot be read or acted on.** Resolving an elementby label is how a caller reaches a link on a page, a picker's `Photos` tab or an autofill sheet's `Fill Strong Password` button. Those elements are no longer in the tree.
- **A caller cannot detect that a modal is up.** The read succeeds and returns a tree; the hosted subtree is simply absent, which is the same answer given when nothing is presented. The two cannot be distinguished.
- While iOS's "Use Strong Password?" sheet is up, the field beneath it keeps only the last keystroke (some weird iOS bug I guess), and typing ten characters on results in one. 

Internally in the companion `--traversal view-hierarchy` still returns the full tree, so nothing is lost from the reader itself it's just that the default no longer reaches it and there is no equivalent for gRPC callers

## Reproducing

Requirements:
- Stock Safari on any booted simulator.
- The SimulatorFrameworkBridge shipped in the companion's `Distribution/Resources`

```sh
GUEST=/path/to/idb/Build/Distribution/Resources/SimulatorFrameworkBridge

# Safari mustn't be running
xcrun simctl terminate booted com.apple.mobilesafari 2>/dev/null

# The app's accessibility server only exists if this is set before it launches.
xcrun simctl spawn booted defaults write com.apple.Accessibility ApplicationAccessibilityEnabled -bool true

# Start mobile safari, open webpage
xcrun simctl launch  booted com.apple.mobilesafari
xcrun simctl openurl booted https://example.com

# Get pid for app.
PID=$(xcrun simctl spawn booted launchctl list | awk '/com.apple.mobilesafari/{print $1}' | head -1)

# The page heading, which is plainly on screen:
xcrun simctl spawn booted "$GUEST" accessibility describe --pid "$PID" | grep -c "Example Domain"
xcrun simctl spawn booted "$GUEST" accessibility describe --pid "$PID" --snapshot-tree 1  | grep -c "Example Domain"
```

Against `main` at `9ca15af`:

```
1     per-node walk
0     single fetch  (--snapshot-tree 1, which is what the default now selects)
```

`--snapshot-tree 1` is what `FBAXBridgeUIAutomation.autoTraversal` sends for a read that names no traversal, so the second command is the behaviour a gRPC caller gets today.

## Where the tree stops

Per-node walk:

```
WKContentView             children=1
  AXRemoteElement           children=1
    WKAccessibilityWebPageObject      children=1
      WebAccessibilityObjectWrapper     children=1
        WebAccessibilityObjectWrapper [Example Domain]  children=3
          ...the rest of the page
```

Single fetch, same screen, same instant:

```
WKContentView             children=1
  AXRemoteElement           children=0
```

The boundary node is present in both. Only its contents are missing.

This matches the mechanism: the single fetch calls `userTestingSnapshotForElement:options:error:` rooted at the application element (`AccessibilityRuntime.m`,
`-snapshotOfElement:attributeNames:namesByNumber:error:`), and that snapshot does not cross into the process drawing the content. The per-node walk asks the accessibility server for each element's children, and the server bridges the boundary.

## Other cases measured

Same build, iPhone 17 Pro, iOS 26.5, whole tree via the `axbridge` backend:

| screen | per-node walk | single fetch (default) |
|---|---|---|
| `PHPickerViewController` presented by a test app | 103 nodes | 17 nodes |
| iOS "Use Strong Password?" autofill sheet | 59 nodes | 40 nodes |
| Settings root (no out-of-process content) | 134 nodes | 134 nodes |

On the autofill sheet the labels absent from the default read are `Close`, `Use Strong Password?`, `Protect your account with a strong, unique password…` and `Fill Strong Password`. A marker read behaves the same way: `AXUniqueId` containing `GenerateStrongPasswordButton` resolves to `Fill Strong Password`
under the walk and answers `the axbridge backend found no element` under the default.

Not everything presented modally is affected — a share sheet in Photos is *drawn in-process* and reads identically both ways 

## Why the equivalence check did not catch it

`27c7821` states:

> Compared node-for-node over every attribute either side emits, the default
> read and `--traversal view-hierarchy` produce the same document.

That holds on a screen with no out-of-process content, and the screens named in the commit (a 167-element Settings screen, an Instagram feed) are such screens — the Settings row above is the same control, identical at 134 nodes both ways. The trees only diverge where something is drawn by another process.

`83566ed`, which added the strategy as opt-in, set the bar this is about:

> This is not the default. […] Label, type and frame match on every screen
> measured, which is evidence but not proof, and a default would need proof.

## Suggested fix

**Removed as wrong** See PR below

## Environment

- idb `main` at `9ca15af`, built locally
- Xcode 26.6, Swift 6.3.3, arm64
- iPhone 17 Pro, iOS 26.5
- macOS 26 (Darwin 25.6.0)

zafnz:
## Motivation

This is a PR to fix #956 . The proposed fix there is wrong, but this fixes it and doesn't lose the advantages from commit 27c7821.  

The primary motivation is that I use idb_companion in [SimGadget](https://github.com/zafnz/simgadget) and until recently it worked fine for navigating, the change traced back to commit id 27c7821 has broken it, specifically:

- Page and sheet content cannot be read or acted on. Resolving an elementby label is how a caller reaches a link on a page, a picker's Photos tab or an autofill sheet's Fill Strong Password button. Those elements are no longer in the tree.
- While iOS's "Use Strong Password?" sheet is up, the field beneath it keeps only the last keystroke (some weird iOS bug I guess), and typing ten characters on results in one.

## Problem
Since that commit, a whole-tree read on the `axbridge` backends returned nothing drawn by another process: a `WKWebView`'s page, a photo picker, the "Use Strong Password?" autofill sheet. The snapshot is served by the process that owns the root element and cannot serialize a subtree another process draws, so the read answered the boundary element with zero children — indistinguishable from nothing being presented. The per-node walk crosses because the accessibility server bridges each walked read. Nothing errored, and the equivalence check behind the default change only measured screens with no out-of-process content.

Each snapshot node carries the element it was read from under `UIAccessibilitySnapshotKeyElement`, and at a boundary that element is owned by the process drawing the subtree. Where a node's nesting is empty and its element's owner differs from the process that served the snapshot, the read now fetches one more snapshot rooted at that element and maps it in place of the stub, recursively, under the same depth and node bounds. A read still costs one fetch per process rather than one round trip per node; a boundary whose owner is not serving keeps the childless stub, which is also the walk's answer there; a boundary a bound stopped reports truncated.

Measured on iOS 26.5 (iPhone 17 Pro), fixed read vs per-node walk, compared node-for-node over every attribute either side emits:

- Safari on example.com: 78 nodes both ways, identical documents, page content present, 2 round trips vs 78.
- PHPickerViewController over a test app: 109 nodes both ways, identical documents, picker content present, 2 round trips vs 109.
- Settings root (no out-of-process content): 134 nodes both ways, identical, still exactly 1 round trip.

The fake runtime now models element ownership so the boundary is expressible: its snapshot stops nesting where the owner changes while its walk still exposes children, which is the shipped behaviour. New tests pin the continuation, the traversal equivalence over a boundary tree, a continuation across a further boundary, the kept stub on a failed continuation, and both bounds holding across a continuation.

## Test Plan

Added unit tests to detect the problem (validated), and verified the fix.

Also tested and validated using the steps outlined in the issue ticket.


lawrencelomax:
lgtm and I think this is the right tradeoff as a default. I did want the default to be as fast as possible and avoid the costs of multiple round trips, but it's unavoidable in this case. rather a slower tree read when remote content is present then completely eliding it.

there might be some conflicting changes since your PR but I hope this is easy to rebase

I might revert some of the opus-isms later on but this is perfectly fine as is.

lawrencelomax:
Also @zafnz I saw from simgadget's readme that you've had issues with the ancient CI builds. This should hopefully be a thing of the past now CI/brew/bottles/releases are back in action over the last few weeks

If that's not working for you then I'd be interested in fixing that.

Issue #894, opened 2025-10-28
video-stream not working on macOS 26 Tahoe


KishanDasani:
I recently updated my macbook air M4 to macOS 26 Tahoe and since then idb stream is broken.

When I connect my iPhone 7 running on v15.8.3 and hit `idb video-stream --udid <device ID>` it gives below output.

"idb video-stream --udid <udid>
Running stream until ^C
2025-10-28 10:22:18,439 [ERROR] - root - Exception thrown in main
Traceback (most recent call last):
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/idb/cli/main.py", line 313, in gen_main
    await root_command.run(args)
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/idb/common/command.py", line 90, in run
    return await self.resolve_command_from_args(args).run(args)
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/idb/cli/__init__.py", line 106, in run
    await self._run_impl(args)
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/idb/cli/__init__.py", line 128, in _run_impl
    await self.run_with_client(args=args, client=client)
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/idb/cli/commands/video.py", line 88, in run_with_client
    async for data in signal_handler_generator(
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/idb/common/signal.py", line 57, in signal_handler_generator
    yield consume.result()
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/idb/grpc/client.py", line 1060, in stream_video
    async for data in iterator:
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/idb/grpc/stream.py", line 51, in generate_bytes
    async for item in stream:
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/grpclib/stream.py", line 61, in __anext__
    message = await self.recv_message()
  File "/Users/admin/.pyenv/versions/3.9.23/lib/python3.9/site-packages/grpclib/client.py", line 426, in recv_message
    await self.recv_initial_metadata()
[...]

Issue #829, opened 2023-04-16
Cannot stream video with mjpeg/minicap format


jsramraj:
### Description

I tried to run the `idb video-stream` command in the terminal. I get output in the terminal only if I use the `h264` format. If I use any other format like `mjpeg` or `minica`, I don't get anything at all in the log. 

### Reproduction

Open the terminal and run the following command
`idb video-stream --fps 30 --format mjpeg --compression-quality 1.0 --udid 0C9CCA92-7D08-4013-889F-F8B31D5F7357`

I get the following thing going in the terminal
`idb video-stream --fps 30 --format mjpeg --compression-quality 1.0 --udid 0C9CCA92-7D08-4013-889F-F8B31D5F7357
Running stream until ÜC`
When I cancel I get the following output
<img width="1838" alt="idb-command" src="https://user-images.githubusercontent.com/1946983/232328648-36b747e6-048b-48a4-aa3c-046cbc3b20db.png">

### Additional Information

* `idb` revision: Latest

PR #958, opened 2026-09-02
Don't pass low-latency rate control to the JPEG encoder

wy0102ylt:
## Summary

`FBSimulatorVideoStreamFramePusher_VideoToolbox.encoderSpecification(for:)` returns
`RequireHardwareAcceleratedVideoEncoder` **plus** `EnableLowLatencyRateControl` for every
format except `.mjpeg(encoder: .allowSoftware)`.

`EnableLowLatencyRateControl` selects the low-latency **video** rate controller. The JPEG
encoder does not accept it, so on macOS 26 `VTCompressionSessionCreate` fails with
`kVTParameterErr` (-12902) and `setup(with:edgeInsets:)` throws
`failedToStartCompressionSession` before a single frame is encoded. Every
`idb video-stream --format mjpeg` and `--format minicap` is therefore dead on macOS 26,
while `--format h264` on the same machine works.

Related: #894, #829. #897 proposes an adjacent change, but against `FBSimulatorVideoStream.m`,
which no longer exists after the Swift rewrite; it also drops the hardware-encoder requirement
altogether, which is not necessary — the rate-control selector alone is what JPEG rejects.

## Fix

Make the mapping exhaustive over `FBVideoStreamFormat` rather than falling through to
"everything else":

- `.mjpeg(encoder: .requireHardware)` and `.minicap` (both JPEG) keep requiring a hardware
  encoder, without the rate-control selector;
- `.mjpeg(encoder: .allowSoftware)` is unchanged;
- `.compressedVideo` keeps both properties — H.264/HEVC behavior is byte-for-byte unchanged;
- `.bgra` returns an empty specification; it takes the bitmap pusher and never creates a
  compression session, so this case is inert and only exists to make the switch exhaustive.

Because the switch is exhaustive, a future format cannot silently inherit the H.264
specification the way the old trailing `return` allowed.

## Test plan

macOS 26 / Xcode 26, arm64, iPhone simulator:

- `idb video-stream --format mjpeg` — before: `VTCompressionSessionCreate` -12902, stream never
  starts; after: session is created and JPEG frames flow.
- `idb video-stream --format minicap` — same before/after.
- `idb video-stream --format h264` and `--format bgra` — unchanged.

This change has been carried in a locally patched companion build (idb v1.5.1 + this diff),
driving a continuous MJPEG framebuffer stream in day-to-day use since 2026-08.

### DIFF
--- a/FBSimulatorControl/Framebuffer/FBSimulatorVideoStreamFramePusher.swift
+++ b/FBSimulatorControl/Framebuffer/FBSimulatorVideoStreamFramePusher.swift
@@ -535,15 +535,26 @@ final class FBSimulatorVideoStreamFramePusher_VideoToolbox: FBSimulatorVideoStre
   }
 
   static func encoderSpecification(for format: FBVideoStreamFormat) -> [String: Any] {
-    if case .mjpeg(encoder: .allowSoftware) = format {
+    switch format {
+    case .mjpeg(encoder: .allowSoftware):
       return [
         kVTVideoEncoderSpecification_EnableHardwareAcceleratedVideoEncoder as String: true
       ]
+    case .mjpeg(encoder: .requireHardware), .minicap:
+      // Low-latency rate control is a video-codec selector. Passing it to the
+      // JPEG encoder makes VTCompressionSessionCreate return kVTParameterErr
+      // (-12902) on macOS 26.
+      return [
+        kVTVideoEncoderSpecification_RequireHardwareAcceleratedVideoEncoder as String: true
+      ]
+    case .compressedVideo:
+      return [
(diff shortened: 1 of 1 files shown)


lawrencelomax:
Thanks for reporting @wy0102ylt

A similar fix is queued within the internal diff D120306658 and will resolve this PR upon internal approval and landing. This PR should then be automatically closed when that is pushed to OSS. I'll look at getting this reviewed and landed within the next few days.

Feel free to reopen if this does not fix the issue or you have any other issues.

lawrencelomax:
A rebase botched the commit that this was in. 

Regardless, this should be closed out by https://github.com/facebook/idb/commit/01caa47b6b794c63f92f0dc570760e71d36c624f

